PR #12, opened 2022-03-10
Improve UI

anhnt4288:
I suggest 2 changes in the UI 
A. First for mobile screen with width <600px. There are two problem: 
1. The navbar is not fit so i decide to hide the button "disconnect"
2. The body from component "Greeter and Storage" is too thin, so i make it wider.
#### Original screen:  
Screen Shot 2022-03-10 at 18 35 59

#### After changes:  
Screen Shot 2022-03-10 at 18 38 16

B. Second , display the error message popup if user reject or others errors
Screen Shot 2022-03-10 at 18 39 32

### DIFF
--- a/packages/react-app/components/ButtonAppBar.tsx
+++ b/packages/react-app/components/ButtonAppBar.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { useEffect, useState } from "react";
 import AppBar from "@mui/material/AppBar";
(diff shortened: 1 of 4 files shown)


critesjosh (approved):
this looks much better, thanks for the contribution!

critesjosh:
@anhnt4288 about how much time did you spend on this to make these improvements?

PR #20, opened 2022-03-23
Added Modifier to Contract

iamoracle:
modifier prevents the storage from being modified outside storage contract.


critesjosh (approved):
👍

PR #21, opened 2022-03-23
Added modifier

iamoracle:
Added an external modifier to ensure the code is only modifiable only outside the contract.


critesjosh (approved):
👍

PR #22, opened 2022-03-26
change imports from relative to absolute

tokoequal:
Changed @celo-progressive-dapp-starter/hardhat imports to absolute as monorepo packages should be configured


critesjosh (approved):
👍

PR #24, opened 2022-03-27
Improve front end architecture

Charlie85270:
### Improve the front end architecture of the project 

- Create Header and Footer components 
- Create reusable Meta Component for the SEO (Will be useful when the app will have more than 1 page) 
- Create AppLayout component to easier organize and have a better view of the layout 
- Fix lint warnings/errors


critesjosh:
thanks for the improvements!

PR #26, opened 2022-03-28
use calldata instead of memory

0xkenta:
The param _greeting will be not updated in setGreeting, so I used calldata instead of memory.

https://ethereum.stackexchange.com/questions/74442/when-should-i-use-calldata-and-when-should-i-use-memory


critesjosh (approved):
nice!

PR #29, opened 2022-03-30
Adding RedStone oracles example

hatskier:
- Added `Oracle.sol` contract
- Added `oracle-test.js` test
- Fixed the `test` command in `package.json`

### DIFF
--- a/packages/hardhat/package.json
+++ b/packages/hardhat/package.json
@@ -6,7 +6,7 @@
     "node": "12.x"
   },
   "scripts": {
-    "test": "hardhat test",
+    "test": "hardhat test --network hardhat",
     "accounts": "hardhat accounts",
     "devchain": "npx celo-devchain --port 8545",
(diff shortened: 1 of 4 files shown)


packages/hardhat/package.json
```diff
@@ -6,7 +6,7 @@
     "node": "12.x"
   },
   "scripts": {
-    "test": "hardhat test",
+    "test": "hardhat test --network hardhat",
```
critesjosh, line 9: Will your tests work on the alfajores testnet? I think the current test scripts run on alfajores by default.

hatskier: Yep, just tested :)

hatskier: I reverted  the `test` script in package.json and added a new one `test-local`. Please let me know if you would prefer to not add it or if you want me to add any other changes to this Pull Request

PR #33, opened 2022-04-01
show actual CELO balance

superbow1:
Added functionality to show users balance in DApp. I think this may be handy in some cases

### DIFF
--- a/packages/react-app/components/AccountInfo.tsx
+++ b/packages/react-app/components/AccountInfo.tsx
@@ -1,0 +1,50 @@
+import * as React from "react";
+import { Divider, Grid, Typography, Link } from "@mui/material";
+
+import { useContractKit } from "@celo-tools/use-contractkit";
+import { useEffect, useState } from "react";
+import { truncateAddress } from "@/utils";
+
+import redstone from "redstone-api"
+
+export function AccountInfo() {
+  const { kit, address, network } = useContractKit();
+  const [balance, setBalance] = useState({ raw: '0', usd: '0' });
+
+  async function fetchBalance() {
+    const { CELO } = await kit.getTotalBalance(address);
+    const celoAmount = kit.web3.utils.fromWei(CELO.toString(), 'ether');
+    const celoUsdPrice = await redstone.getPrice('CELO');
+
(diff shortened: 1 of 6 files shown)


critesjosh:
Thanks for the PR! I like the idea, I think the UI could be cleaned up a bit. It looks pretty wonky a mobile device as it is.

Screen Shot 2022-04-01 at 2 22 57 PM

Maybe truncate the balance so it shows only 2 decimal place so the chip is smaller. And/or update the account chip so when you click on it, it shows additional information like the network details and account balance. Or maybe even add an "Account" tab similar to the "Storage" and "Greeter" tabs, but that shows account details. I am also open to suggestions.

superbow1:
> Thanks for the PR! I like the idea, I think the UI could be cleaned up a bit. It looks pretty wonky a mobile device as it is.
> 
> Screen Shot 2022-04-01 at 2 22 57 PM
> 
> Maybe truncate the balance so it shows only 2 decimal place so the chip is smaller. And/or update the account chip so when you click on it, it shows additional information like the network details and account balance. Or maybe even add an "Account" tab similar to the "Storage" and "Greeter" tabs, but that shows account details. I am also open to suggestions.

Adding "Account" tab is a good idea, I will update my PR

critesjosh:
Cool! Since there a bunch more space now, it might be cool to add cUSD and cEUR balances as well. wdyt?

Also, I am going to shorten the tab names because currently only 2 will appear on a mobile device at a time.

superbow1:
> Cool! Since there a bunch more space now, it might be cool to add cUSD and cEUR balances as well. wdyt?
> 
> Also, I am going to shorten the tab names because currently only 2 will appear on a mobile device at a time.

Added cEUR and cUSD

superbow1:
> Also, I am going to shorten the tab names because currently only 2 will appear on a mobile device at a time.

I think it is better to make tabs scrollable. It just needs few additional props

https://user-images.githubusercontent.com/102758523/161336509-fbdd79be-ff45-466e-8cd6-05db5dabf543.mp4

critesjosh:
looks great, thanks!

PR #37, opened 2022-05-02
Create Eth_converter.sol

darom1998:
It's an eth converter

### DIFF
--- a/Eth_converter.sol
+++ b/Eth_converter.sol
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.7;
+
+import "@chainlink/contracts/src/v0.8/interfaces/AggregatorV3Interface.sol";
+
+contract PriceConsumerV3 {
+
+    AggregatorV3Interface internal priceFeed_eth_usd;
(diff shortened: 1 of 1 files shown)


critesjosh:
Chainlink isn't on Celo yet and I don't see anything at the account addresses in the contract.

darom1998:
Hi those addresses are for the api between my contract and the pricefeed
that exist outside the blockchain , and for chainlink it's crucial and it's
solving numerous problems.
Can you precise the address that you are looking for and thanks.

Eth_converter.sol
```diff
@@ -0,0 +1,51 @@
...
+
+import "@chainlink/contracts/src/v0.8/interfaces/AggregatorV3Interface.sol";
+
+contract PriceConsumerV3 {
+
+    AggregatorV3Interface internal priceFeed_eth_usd;
+    AggregatorV3Interface internal priceFeed_btc_eth;
+
+    /**
+     * Network: Kovan
```
critesjosh, line 12: as noted here, these contract addresses are for the Kovan Ethereum testnet. This isn't relevant for Celo.

PR #46, opened 2022-06-19
changed react version to 17.0.2 to match the react version of the react app template

therealharpaljadeja:
React Native also uses react under the hood and having different versions is giving Invalid Hook call error.

More on this here

https://reactjs.org/warnings/invalid-hook-call-warning.html


joenyzio:
LGTM!

Issue #55, opened 2022-07-03
React-Celo Integration


joenyzio:
Composer currently uses use-contractkit. Update code to use latest version - react-celo.

import { useContractKit } from "@celo-tools/use-contractkit";

joenyzio:
Done with https://github.com/celo-org/celo-composer/pull/102

PR #102, opened 2022-07-05
React-Celo Integration

leetdev:
Migrate web app from `contract-kit` to `react-celo` as per #55.


leetdev:
I've managed to fix the broken deploy - turns out the `.nvmrc` file was removed in a previous commit, which caused Netlify to use the wrong version of Node.

joenyzio:
Awesome PR LGTM!

Issue #106, opened 2022-07-18
Error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported


alofeoluwafemi:
<img width="1440" alt="Screen Shot 2022-07-18 at 7 21 08 PM" src="https://user-images.githubusercontent.com/7295729/179567959-e79198ed-d2f2-4e6c-add2-770038b758db.png">

When connecting a wallet with low Celo balance everything seem to work fine until you open a Celo wallet with a high balance in my case 3700+ Celo Tokens. Converting the `CELO` variable from `getTotalBalnce` returns an exponential string which `web3.utils.fromWei` cannot handle.

PR #107, opened 2022-07-18
Fix: For issue #106

alofeoluwafemi:
Fix for Issue: https://github.com/celo-org/celo-composer/issues/106

A large wallet balance cannot be displayed due to the error `number-to-bn` from the big number library. It expects string such as `3.12344` but get exponential string such as `3.112345671829082654e21`, the EVM returns above MAX number javascript can handle. Using `toFixed` of big number library fixes the issue and additionally used ethers js for Parsing type of `BigNumberish` to ether value as a string.


slim12kg:
ERC20 Wallet 0x3472059945ee170660a9A97892a3cf77857Eba3A

joenyzio (approved):

joenyzio (approved):

PR #112, opened 2022-07-26
Fix error : validateDOMNesting(...): <div> cannot appear as a descendant of <p>

slim12kg:
Clear error validateDOMNesting(...): <div> cannot appear as a descendant of <p> and made changes to Celo network account config in hardhat to accept array and not string


joenyzio (approved):
LGTM!

PR #115, opened 2022-08-07
Feature dev tools

alofeoluwafemi:
Add and configure dev tools, such as solidity linter using sollint, prettier for formatter also configured to work with sollint and husky to run pre-commit hooks. Other pre-commit hooks for react and typescript can be later configured


nestorbonilla:
Closing PR for outdated.

PR #118, opened 2022-08-14
Fix error caused by merge conflict

leetdev:
Merging a previous PR seems to have introduced an error in the `AccountInfo` component that was likely caused by an incorrect resolution of a merge conflict. Since it prevents the React App from loading, I've introduced a fix here.


joenyzio (approved):

joenyzio:
LGTM!

Issue #69, opened 2022-07-03
Composer for React Native without Expo


joenyzio:
Add option to use Composer for React Native without Expo.

nestorbonilla:
It seems the issue has been solved.
Closing issue.

PR #121, opened 2022-08-28
added support for react native app without expo

Mujhtech:
Added option to use Composer for React Native without Expo.


Mujhtech:
Fixed #69 .

PR #125, opened 2022-09-19
Added query intent for android 11+ users to be able to open a link

Mujhtech:
For both flutter app & react native app without expo


nestorbonilla:
This should be fixed now; if not, feel free to open it again.

PR #126, opened 2022-09-21
Celo CLI Support added

viral-sangani:
Added CLI support for Celo Composer. User can use `npx` to setup celo starter kit for different packages.
@joenyzio


joenyzio (approved):
LGTM!

PR #135, opened 2022-10-30
Cleanup: delete temporary OS files (.DS_Store)

jodaAtGithub:
A `.DS_Store` file is specific temporary local file, generated by Mac OSX, containing specific settings for a local setup. It is common practice to prevent them from pushing into the VCS. 

The `.DS_Store` filename is already listed in `.gitignore`, I guess that these files was already listed before they got ignored.


jodaAtGithub:
Hm, how did that change broke the deployment? Would be awesome if you could give me some guidance to solve that.

nestorbonilla (approved):
LGTM

PR #136, opened 2022-10-31
celo-composer updates v1

nestorbonilla:
Updates:
- celo-composer CLI script => update to clone only selected projects, but on different folders inside packages, rather than one single folder.
- react-app => libraries, new contract path, and style updates.
- hardhat => libraries, solidity version standardization updates, and hardhat-celo verification option integrated.


viral-sangani:
@nestorbonilla Is there any way we can suppress the `echo` output printed in the CLI command? I tried looking but couldn't find a solution. I guess the library does not support it natively. We might have to look for other packages for that echo command because printing the while package.json twice in the terminal is not a good experience.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celo/celo-composer",
-  "version": "0.0.2-alpha",
+  "version": "0.0.3",
```
viral-sangani, line 3: We can make the version `1.0.1` as the current live version on npm is `1.0.0`.

bin/create.js
```diff
@@ -9,75 +8,234 @@ const { ensureDir, readdir } = require("fs-extra");
...
 
-    const { stdout, stderr, code } = shell.exec(
-      `git sparse-checkout set packages/${slug} packages/hardhat`,
-      {
-        silent: true,
+      // clone to local only the projects user wants
+      const { stdout, stderr, code } = shell.exec(
+        `git sparse-checkout add packages/${selectedPackages[x]}`, { silent: true }
+      );
+
```
viral-sangani, line 177: Need to add a check `selectedPackages[x] != selectedPackages[3]` (i.e. flutter-app).
@nestorbonilla

nestorbonilla: done.

package.json
nestorbonilla: done.

bin/create.js
```diff
@@ -9,75 +8,234 @@ const { ensureDir, readdir } = require("fs-extra");
...
+      }
+      
+      if (selectedPackages[x] == availablePackages[7].name) {
+        packageJson.scripts["subgraph-get-abi"] = `yarn workspace @${projectName}/subgraphs get-abi`;
+      }
+      
+    }
+
+    shell.echo(JSON.stringify(packageJson, "", 4)).to("package.json");
+
```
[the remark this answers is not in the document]
nestorbonilla: done.

viral-sangani (approved):

PR #150, opened 2022-11-09
fix: edit `name` in package.json

therealharpaljadeja:
Change the name in every package generated in order to match with the root package.json


nestorbonilla (approved):
LGTM

PR #154, opened 2022-11-14
refactor: allow CLI to receive answers programmatically

therealharpaljadeja:
Now the CLI can receive answers to the questions programmatically, this helps in testing the CLI scripts.


nestorbonilla:
@therealharpaljadeja do you want to update it with current updates, or should we close it?

PR #166, opened 2022-12-19
cli script and react-app update

nestorbonilla:
**Background**
The previous react-app package did not fall into the template category, as it had too many components and third-party libraries, and some developers expressed that many times they had to remove many components to use the package.

Therefore, the previous react-app package has been moved to an [example project](https://github.com/celo-example/celo-composer-react-example) using celo-composer.

**Improvements**
The CLI script has been updated together with the new react-app package, and now devs can select their preferred web3 front-end library, having as options:

- rainbowkit-celo
- react-celo


viral-sangani:
@nestorbonilla I think `@celo/react-celo` is missing in the `package.json`. Getting this error - 
```
Type error: Cannot find module '@celo/react-celo' or its corresponding type declarations.
```

viral-sangani (approved):

Issue #51, opened 2022-07-03
ERC20 Example


joenyzio:
Add basic [ERC20](https://docs.openzeppelin.com/contracts/4.x/erc20) token functionality

nikbhintade:
Hi @joenyzio, I saw that there are lot of issues to add examples. Can you tell me what is the contribution process for adding these examples? If all these issues (add example of X) are active then I want to complete as many issues as possible.

nikbhintade:
I want to complete these as a showcase for my development skills which would help me in the career change.

therealharpaljadeja:
@nikbhintade you still interested?

PR #190, opened 2023-03-24
Add SimpleERC20 contract

Ankush263:
Add SimpleERC20 smartcontract #51


Ankush263:
Should I have to change something to my code??

PR #199, opened 2023-05-30
Add WalletConnect v2 support for React Native Expo example

therealharpaljadeja:


packages/react-native-app/screens/Account.tsx
```diff
@@ -32,11 +30,11 @@ export default function Account() {
                 </Button>
             </View>
             <View style={styles.separator}></View>
-            <Button onPress={() => connector.killSession()}>
+            {/* <Button onPress={() => connector.killSession()}>
```
ignaciosantise, line 33: You can use the provider to disconnect

`const { provider } = useWeb3Modal();`

`<Button onPress={() => provider?.disconnect()}>`

Ernesto-tha-great (approved):

packages/react-native-app/screens/Account.tsx
therealharpaljadeja: Yeah I was wondering the same, how to disconnect thanks for answering it here!

viral-sangani (approved):

viral-sangani (approved):

PR #254, opened 2023-11-01
updated react-app to nextjs 14

viral-sangani:
next version updated to 14.


therealharpaljadeja:
Don't we need to change the file structure?

PR #283, opened 2024-03-22
feat: `celo-composer` `v.1.0.13`

therealharpaljadeja:
Celo Composer support for latest `viem` and `wagmi` libraries

- `viem` (2.8.18)
- `wagmi` (2.5.12)
- `rainbowkit-celo` (removed)
- `@rainbow-me/rainbowkit` (2.0.2)

@viral-sangani @GigaHierz Test thoroughly!

Hackers can now build with Metamask on their pc and then later demo it in Site tester.

### DIFF
--- a/packages/react-app/pages/_app.tsx
+++ b/packages/react-app/pages/_app.tsx
@@ -1,45 +1,53 @@
-import { Alfajores, Celo } from "@celo/rainbowkit-celo/chains";
-import celoGroups from "@celo/rainbowkit-celo/lists";
-import { RainbowKitProvider } from "@rainbow-me/rainbowkit";
+import {
+    RainbowKitProvider,
+    connectorsForWallets,
+} from "@rainbow-me/rainbowkit";
+import { injectedWallet } from "@rainbow-me/rainbowkit/wallets";
 import "@rainbow-me/rainbowkit/styles.css";
 import type { AppProps } from "next/app";
-import { WagmiConfig, configureChains, createConfig } from "wagmi";
-import { publicProvider } from "wagmi/providers/public";
+import { http, WagmiProvider, createConfig } from "wagmi";
 import Layout from "../components/Layout";
 import "../styles/globals.css";
+import { celo, celoAlfajores } from "wagmi/chains";
 
-const projectId = process.env.NEXT_PUBLIC_WC_PROJECT_ID as string; // get one at https://cloud.walletconnect.com/app
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 
-const { chains, publicClient } = configureChains(
-  [Celo, Alfajores],
-  [publicProvider()]
+const connectors = connectorsForWallets(
+    [
+        {
+            groupName: "Recommended",
+            wallets: [injectedWallet],
+        },
+    ],
+    {
+        appName: "Celo Composer",
+        projectId: "044601f65212332475a09bc14ceb3c34",
+    }
 );
 
-const connectors = celoGroups({
-  chains,
-  projectId,
-  appName: (typeof document === "object" && document.title) || "Your App Name",
+const config = createConfig({
+    connectors,
+    chains: [celo, celoAlfajores],
+    transports: {
+        [celo.id]: http(),
+        [celoAlfajores.id]: http(),
+    },
 });
 
-const appInfo = {
-  appName: "Celo Composer",
-};
-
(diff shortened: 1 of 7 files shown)


packages/react-app/pages/_app.tsx
```diff
@@ -1,45 +1,53 @@
...
+const connectors = connectorsForWallets(
+    [
+        {
+            groupName: "Recommended",
+            wallets: [injectedWallet],
+        },
+    ],
+    {
+        appName: "Celo Composer",
+        projectId: "044601f65212332475a09bc14ceb3c34",
```
GigaHierz, line 24: `projectId: "044601f65212332475a09bc14ceb3c34" || process.env.NEXT_PUBLIC_WC_PROJECT_ID as string`

If we just have the hardcoded variable people might miss that they have to exchange this for production.

packages/react-app/pages/index.tsx
```diff
@@ -2,23 +2,36 @@ import { useEffect, useState } from "react";
...
-      </div>
-      {isConnected && (
-        <div className="h2 text-center">Your address: {userAddress}</div>
-      )}
-    </div>
-  );
+    return (
+        <div className="flex flex-col justify-center items-center">
+            <div className="h1">
+                There you go... a canvas for your next Celo project!
```
GigaHierz, line 26: We talked about maybe having a similar setup as the minipay template here so people already have a read and a write function, that they can just copy. Could be something as easy as readign their cUSD balance and sending cUSD

packages/react-app/pages/_app.tsx
therealharpaljadeja: Rather than throwing an error right off the bat, this fix lets the "beginners" get to the main screen without any errors.

Production-level developers will know to change project id, we also never recommended to use celo-composer in production environment.

packages/react-app/pages/index.tsx
therealharpaljadeja: Yeah that can be done, will add that in separate PR.

But did you test celo-composer and did it work for you (without any errors and having to change anything)?

GigaHierz:
okay, so we are not using wallet connect here right? it's not shown to me in the options. 

image

GigaHierz:
when installing with yarn I am not getting any issues but installing dependencies with `npm i` leads to this error:

image

therealharpaljadeja:
> okay, so we are not using wallet connect here right? it's not shown to me in the options.
> 
> image

Yeah at hackathons, people use browser-based wallets. Also, Walletconnect shows a ton of wallet options which may not support Celo and if someone tries that wallet, unexpected errors can pop up...

packages/react-app/pages/_app.tsx
therealharpaljadeja: @GigaHierz are we good on this?

packages/react-app/pages/index.tsx
therealharpaljadeja: Separate PR for that coming soon...

packages/react-app/pages/_app.tsx
GigaHierz: only concern I once had a project failing at one point because of having had that number in public for many people to use. So, I think. yes let's add it but also ask people to add their number in the .env file. if they don't they can still find this easier as the source of error.

therealharpaljadeja: > my only concern is that I once had a project failing at one point because of having had that number in public for many people to use. So, I think. yes let's add it but also ask people to add their number in the .env file. if they don't they can still find this easier as the source of error.

Failing because of `projectId` is strange, is it possible to have a look at the example that failed? Can you potentially recreate that error if possible?

viral-sangani:
Tried and tested in - 
- Browser
- Incognito Mode
- Minipay

LGTM

packages/react-app/pages/_app.tsx
therealharpaljadeja: https://celofoundation.slack.com/archives/C01V5UXJ6RY/p1687946055310059?thread_ts=1687551395.330579&cid=C01V5UXJ6RY

GigaHierz: I can't find it. looked through all my former projectIDs and I used them in many examples and tutorials and they are all working. So, tbh. it should be fine.

GigaHierz (approved):

PR #307, opened 2024-10-24
feat: migrate app router

ankit-tailor:


viral-sangani:
@ankit-tailor 
can you removed 
`swcMinify` from nextjs config?

Issue #297, opened 2024-10-02
[Feature]: Add support for Next.js App Router Template


Philix27:
### Tell us about your idea

- The cli app always generates a Next.js web app which uses the page router, and  since the coming of Next.js v14, app router has been the go to standard due to its supports server components. 

- Expected outcome: To provide a template for Next.js app router with support for all configurations as the page router. 

### Anything else?
Link to an example repo of Next.js App Router for celo composer. [repository](https://github.com/Philix27/celo-composer-nextjs-approuter)

_No response_

PR #310, opened 2024-10-24
Add support for Sveltekit + Wagmi template

Philix27:
This PR address this issue. #297

Because the CLI is designed to clone a repository and make modifications where necessary, the template currenly sits on my personal repository. The generated code for all templates is not onthis repository.

I can always push  it to Celo Org when a repository has been created for it.  

celo-next-router


viral-sangani:
@Philix27 Thanks for the PR but we are currently moving from page router to app router in main react-app. That will be live by tomorrow.

PR #320, opened 2024-11-06
Support for Sveltekit Template

### LINKED ISSUE #313: [Feature]: Add support for Sveltekit
### Tell us about your idea

Celo composer to support Sveltekit base template configured for development on the Celo blockchain.

Expected Outcome
- Cli generates code with Svelkit + Wagmi + WalletConnect template

### Anything else?

_No response_

Philix27:
This PR resolves #313. 

Features
- Wallet connection using web3Modal
- Interact wallet/account details using wagmi core
- Interact with Smart contact using wagmi core


viral-sangani:
Hey @Philix27 can you please also add Sveltekit code in this PR as well instead of linking it to your repo?
Sorry for late reply.

Philix27:
I have moved the template to this repository.  @viral-sangani 

Socket Security is pointing to sveltekit core packages (sveltejs/kit, esbuild) needed to run the app. I feel we can let some of  these packages pass through.

viral-sangani:
@Philix27 It does make sense to clone the sub directory rather than cloning from different URL, but we are trying to keep all the templates outside the celo composer repo and all the code components i.e. Basic React + Hardhat etc in the main repo. 
How about this - 
- I have create a new repo - [https://github.com/celo-org/svelte-template](https://github.com/celo-org/svelte-template), if you move your current svelte code to this repo, and use this as clone URL that would be best.
- Also remove the svelte code from here.

Initially we planned to keep the code here and give an option between ReactJS and Svelte, but looking at the requirement for svelte vs react in web3 development, we want to keep react as default options and svelte as a template which is maintained by community.

Philix27:
@viral-sangani  Okay, I tried pushing to the new repo but my access was denied.

viral-sangani:
@Philix27 Can you try now? You can fork the repo and then add your code and create PR to upstream.

Philix27:
@viral-sangani Okay, the Sveltekit template has been removed and a PR created.

PR #332, opened 2025-02-12
update dependency and add use client to Header

GigaHierz:


sodofi (approved):

sodofi (changes requested):
- Not passing Socket Security: Pull Request Alerts

PR #359, opened 2025-07-27
fix(create): Resolve ERR_REQUIRE_ESM error for chalk/ora 

### LINKED ISSUE #358: [Bug]: Error: require() of ES Module chalk not supported in @celo/celo-composer
### Which framework are you using?

React

### What happened?

When running npx @celo/celo-composer@latest create on Node.js v18 or v20, I get the following error:
`Error [ERR_REQUIRE_ESM]: require() of ES Module .../chalk/source/index.js ... not supported.
Instead change the require of index.js ... to a dynamic import() which is available in all CommonJS modules.
`

### Steps to reproduce

1. Install Node.js v18 or v20

2. Run: npx @celo/celo-composer@latest create

Expected Behavior:
The command should run without errors and allow me to scaffold a new Celo project.

Actual Behavior:
The command fails with an ESM import error related to chalk.

Environment:

Node.js version: v18.20.8 and v20.x
npm version: v10.8.2
OS: [Your OS, e.g., macOS 14.4.1]
Additional Context:
This appears to be due to chalk being ESM-only, but @celo/celo-composer is using require('chalk') instead of a dynamic import().

### Anything else?

`npx @celo/celo-composer@latest create
/Users/la/.npm/_npx/5028351eb6129d71/node_modules/@celo/celo-composer/dist/commands/create.js:7
const chalk_1 = __importDefault(require("chalk"));
                                ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/la/.npm/_npx/5028351eb6129d71/node_modules/chalk/source/index.js from /Users/la/.npm/_npx/5028351eb6129d71/node_modules/@celo/celo-composer/dist/commands/create.js not supported.
Instead change the require of index.js in /Users/la/.npm/_npx/5028351eb6129d71/node_modules/@celo/celo-composer/dist/commands/create.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/la/.npm/_npx/5028351eb6129d71/node_modules/@celo/celo-composer/dist/commands/create.js:7:33) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.12.0`

SabaliDev:
The create command was failing with an ERR_REQUIRE_ESM error when executed via npx. This issue stemmed from the use of static require() imports for chalk and ora, both of which are now published as pure ES Modules.

This commit refactors the code to use dynamic import() for these dependencies, which is the correct and compatible approach when loading ESM packages in a CommonJS context. This resolves the runtime error and restores functionality when running the CLI via npx.

Fixes #358

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     }
   },
   "scripts": {
+    "dev": "oclif dev",
     "build": "shx rm -rf dist && tsc -b",
     "lint": "eslint . --ext .ts",
     "postpack": "shx rm -f oclif.manifest.json",
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -1,7 +1,6 @@
 /* eslint-disable perfectionist/sort-imports */
 /* eslint-disable no-await-in-loop */
 import { Command, Flags } from "@oclif/core";
-import chalk from "chalk";
 import { execa } from "execa";
 import fs from "fs-extra";
 import inquirer from "inquirer";
@@ -9,7 +8,6 @@ import kebabCase from "lodash.kebabcase";
 import emoji from "node-emoji";
 import { readdir } from "node:fs/promises";
 import path from "node:path";
-import ora from "ora";
 
 import {
   BASE_URL,
@@ -65,6 +63,7 @@ export default class Create extends Command {
   };
 
   public async run(): Promise<void> {
+    const { default: chalk } = await import("chalk");
     const { flags } = await this.parse(Create);
 
     const packages = ["react-app"];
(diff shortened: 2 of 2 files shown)


GigaHierz:
Thanks for digging into this and for the workaround! The underlying issue was resolved a different way: in #364 the whole CLI was migrated to native ESM (`"type": "module"`), so chalk 5 and ora 7 are now imported statically without any ERR_REQUIRE_ESM problems, and `src/commands/create.ts` has since been rewritten entirely (the oclif-based version this PR patches no longer exists). The CLI on current main builds and runs cleanly, with a smoke-test suite covering the scaffolding paths. Recommending closure as superseded — if you hit any ESM issue on the current `@celo/celo-composer@latest`, please open a fresh issue and we'll jump on it.

PR #380, opened 2025-12-18
chore: add deployment guide and CI/CD workflows for npm publishing

viral-sangani:


GigaHierz (approved):

GigaHierz:
This is still very much wanted — the repo currently has no CI workflows at all — and it composes well with what landed this week: `pnpm test` now runs a real scaffold smoke suite (#447), and `scripts/release.sh` already pushes the tag your publish workflow would pick up. Before merging it needs a refresh:

1. **Step-ordering bug in all three workflows:** `actions/setup-node` with `cache: 'pnpm'` runs before `pnpm/action-setup` — the pnpm cache requires pnpm on PATH, so every job dies at setup as written. Move `pnpm/action-setup` first in `ci.yml`, `publish.yml`, `publish-beta.yml`.
2. **Drop the `package.json` 2.4.13 → 2.4.14 bump** — versioning stays with `release.sh`.
3. **Reconcile `scripts/release.sh` in the same PR:** after #448 it explicitly says nothing publishes automatically, which stops being true the moment these workflows land. Its epilogue should describe the tag-triggered publish, and `DEPLOYMENT.md` should name `release.sh` as the entry point.
4. Pin Node to an LTS (22/24) rather than 26, or a matrix over `engines: >=18`.
5. `npm uninstall -g celo-composer` in ci.yml → `@celo/celo-composer` (the `|| true` currently masks the wrong name).
6. Add a least-privilege `permissions:` block to ci.yml, pin actions by SHA, and consider a `concurrency` group + `npm publish --provenance` — the repo's own StepSecurity checks will flag these.
7. In `publish.yml`, the manual-dispatch path publishes a version bump it never commits back — simplest is removing the `workflow_dispatch` version input and making tags the only trigger.

The tarball glob, pre-release tag gating, and the pack/install/`--help` smoke check all verified correct — happy to re-approve and merge quickly with the above addressed.

GigaHierz:
Refreshed as a maintainer edit (rebased onto main + one commit) addressing all seven points from the review: pnpm/action-setup now runs before setup-node in all three workflows (and is bumped to v6, SHA-pinned along with checkout/setup-node), Node is on 22 LTS, the package.json bump is dropped, publish workflows are tag-triggered only with a new tag-vs-package.json verification step and npm provenance, ci.yml has least-privilege permissions + a concurrency group, the global-uninstall step uses the real package name, and release.sh/DEPLOYMENT.md now describe the automated flow consistently (release.sh is the entry point; its 'nothing publishes automatically' epilogue is replaced). One operational prerequisite before the first tagged release: provision NPM_TOKEN in the repo secrets.

GigaHierz (approved):
Re-approving the refreshed version: all three workflows parse, the setup-ordering bug is fixed, versioning stays in the repo, and the publish gate now runs the real smoke-test suite via pnpm test. The CI workflow will start protecting PRs the moment this merges.

PR #381, opened 2026-03-17
Report the package.json version number

m-chrzan:
The output of the `--version` flag had been hardcoded to `1.0.0`. With this change, the version number is read from package.json.


GigaHierz:
Thanks — this was exactly the right fix, and the identical mechanism (`createRequire` reading package.json at runtime) landed on main via #439, so `--version` now reports the real package version. Recommending closure as already-fixed rather than asking for a rebase. Appreciated!

PR #388, opened 2026-06-12
feat: add stablecoin gas abstraction helper and transaction demo for Farcaster miniapps

samuelchimmy:
This PR adds support for Celo's native gas abstraction feature in the Farcaster Miniapp (Frame v2) template. It includes a custom stablecoin gas helper (@/lib/celo-gas) that maps mainnet/Sepolia fee currency addresses (cUSD, cEUR, cREAL, USDT) and an interactive transaction demo card in page.tsx where developers can test sending transactions using their preferred fee currency. It also adds documentation to FARCASTER_SETUP.md and farcaster-miniapp.mdx.

### DIFF
--- a/docs/templates/farcaster-miniapp.mdx
+++ b/docs/templates/farcaster-miniapp.mdx
@@ -35,3 +35,29 @@ To make your Miniapp discoverable, you must sign a manifest with your Farcaster 
   `https://farcaster.xyz/~/developers/mini-apps/manifest?domain=YOUR_DOMAIN` to
   sign the manifest.
 </Info>
+
+## Celo Gas Abstraction (Fee Currency)
+
+Celo's unique native gas abstraction allows users of your Farcaster Miniapp to pay gas fees in stablecoins (`cUSD`, `cEUR`, `cREAL`, `USDT`) instead of native `CELO` tokens.
+
+The template contains a predefined helper file:
+- **`apps/web/src/lib/celo-gas.ts`**: Contains fee currency contract mappings for Celo Mainnet and Celo Sepolia testnet, along with the `getFeeCurrencyAddress` function.
+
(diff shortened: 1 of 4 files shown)


GigaHierz:
Thanks — fee abstraction is exactly the kind of Celo-native feature we want the Farcaster template to showcase, so we'd love a refreshed version of this. It needs rework before it can merge:

1. **Rebase onto current main** — the template migrated to the `@farcaster/miniapp-*` SDKs this week (#449), so the `@farcaster/frame-sdk` references and "Frame v2" framing are outdated, and the page/component structure changed.
2. **The fee-currency address table must be corrected against chain state:** the mainnet cEUR and cREAL entries are not the real token addresses (real: cEUR `0xD8763CBa276a3738E6DE85b4b3bF5FDed6D6cA73`, cREAL `0xe8537a3d056DA446677B9E9d6c5dB704EaAb4787`); USDT (6 decimals) must go through its FeeCurrencyAdapter (`0x0E2A3e05bc9A16F5292A6170456A710cb89C6f72`) rather than the token address; and the "celoSepolia" block currently contains Alfajores addresses (partly incorrect even for Alfajores) — on Celo Sepolia the on-chain-verified USDm is `0xdE9e4C3ce781b4bA68120d6261cbad65ce0aB00b`.
3. **Naming should follow the on-chain rebrand** — the mainnet cUSD contract now reports `USDm` (see the note in `templates/minipay/components/user-balance.tsx.hbs`), and only tokens actually whitelisted as fee currencies per network should be listed.

If you refresh along those lines we'd be glad to review again promptly.

GigaHierz:
Closing per maintainer decision — the fee-abstraction idea is appreciated, but with the corrected scope (address table needs full rework against verified chain state, and the template has since migrated SDKs) this isn't something we'll pursue in this form. The refresh guidance in the comment above stands if anyone wants to pick the feature up fresh.

PR #393, opened 2026-08-09
feat(templates): add an x402 paid API template

### LINKED ISSUE #392: feat: add an x402 template — scaffold an API that agents can pay for
## Summary

Add an `x402` template so `create` can scaffold a service that charges per request and gets paid in stablecoins on Celo, through the hosted facilitator at `api.x402.celo.org`.

There is no x402 anywhere in this repo today, and Celo runs a facilitator that anyone can point at. Someone building a paid API for agents currently has to assemble the seller by hand, work out where the payment requirements go, what belongs in `extra`, and why the price is in raw units. That is the content of a template.

## Why it fits here

Composer already scaffolds the two Celo-specific things a developer would otherwise get wrong on their own: MiniPay and Farcaster. x402 is the same shape of problem. The protocol is generic; the parts that are easy to get wrong are Celo-specific — which facilitator, which asset, and the EIP-712 domain each token signs with.

The part developers do not expect, and the reason it deserves a scaffold rather than a snippet: **the buyer needs no CELO.** The authorization is signed off-chain and the facilitator sponsors the gas, so an agent can hold nothing but a stablecoin and still transact.

## Proposed shape

Three wiring points, matching how `minipay` and `ai-chat` are done today:

| where | change |
|---|---|
| `templates/x402/` | new directory of `.hbs` files |
| `src/commands/create.ts` | one entry in the `templateType` choices array |
| `src/plopfile.ts` | actions gated on `data.templateType === "x402"` |

## What it generates

- A seller route that returns `402` with payment requirements and serves the paid response once payment verifies
- A buyer/agent script that reads the header back and pays, so the whole loop is runnable from one `create`
- Defaults to Celo Sepolia and USDC, so it costs nothing to try before anyone touches mainnet

## Open questions for maintainers

1. **Wallet prompt.** `farcaster-miniapp`, `minipay` and `ai-chat` skip the `walletProvider` question. The seller side needs no wallet, but the buyer side needs a signer. Skip it, or keep it for the buyer?
2. **Scope.** Seller only, or seller plus the buyer script? Seller alone is smaller; the pair is what makes it demonstrable without extra docs.
3. **Contracts.** Nothing on-chain is needed for this, so presumably it skips the `contracts` prompt entirely. Confirm?

I have a live x402 seller on Celo mainnet that I built without a template, so I can bring the parts that cost me time rather than guess at them. Happy to open the PR once the shape is agreed.

ghozzza:
Scaffolds an API that agents can pay for in stablecoins on Celo, as a
new choice in the create flow:

  npx @celo/celo-composer create my-app -t x402

The template writes apps/api, a second workspace package holding both
halves of the payment: a Hono seller whose POST /paid answers 402 with
what it accepts, and a buyer script that reads those requirements, pays
them, and prints the settlement hash. It points at Celo Sepolia by
default so the whole loop runs for free.

Closes #392.

Why apps/api rather than a route in apps/web
--------------------------------------------
@x402/next peer-depends on next >= 16.2.6 and the base template pins
next ^14.0.0. package.json.hbs already varies dependencies per template,
so Next 16 could be pinned for x402 alone — but that would run the whole
base web app (layout, page, the ui components, eslint-config-next@14) on
a major version it has never been tested against, for a seller that has
no reason to live in the frontend. @x402/hono has no framework
constraint, pnpm-workspace.yaml already globs apps/*, and apps/web is
left byte-identical to every other template. Verified: the generated
apps/web still builds.

Both @x402 server adapters also peer-depend on @x402/paywall, which
serves a browser paywall page and pulls in Solana and Algorand wallet
stacks plus React 19. An API seller never renders that page, so the
generated root package.json declares it under pnpm.peerDependencyRules
.ignoreMissing. Verified after a real install: 444 packages, of which
zero match paywall, solana or algorand.

Three defaults, all reversible
------------------------------
The issue left three scope questions open. Rather than wait, this picks
defaults and states them so they are visible on first read:

- the walletProvider prompt is skipped — the seller verifies signatures
  rather than holding a wallet, and the buyer reads a key from env
- both halves ship, because the loop is the point of a scaffold
- the contracts prompt is skipped — nothing is deployed

Adding a template is twelve edits, not one
------------------------------------------
The issue described this as a directory plus one line in the choices
array. Counted in the diff, a template type has to be threaded through
twelve places: the picker, the walletProvider prompt's when(), the
contractFramework prompt's when(), two branches of the non-interactive
--yes path, two final fallbacks, and the -t help string in index.ts —
plus four in plopfile.ts, two skip guards and two actions. Missing one
does not break the CLI; it just quietly asks for a wallet provider the
template has no use for.

So src/generators/template-wiring.test.ts generates real projects and
asserts on the output. It is the first test under src/ in this repo's
history — jest was configured with --passWithNoTests and `git log --all`
shows no test file has ever existed there. It drives the CLI as a
subprocess through tsx, because the source is ESM and several
dependencies are ESM-only, which Jest cannot load in-process without
converting the whole runner. Converting it is a separate change.

ESLint's type-aware rules need a project that covers the test file,
while the build tsconfig excludes *.test.ts so it never reaches dist/.
Hence tsconfig.eslint.json.

Verification
------------
Generated a project, installed it, and ran it. POST /paid with no
payment header returns 402 with a payment-required header decoding to
x402Version 2, network eip155:11142220, asset
0x01C5C0122039549AD1493B8220cABEdD739BC44E, amount 10000, and the
EIP-712 domain. Both facilitator URLs were checked live beforehand and
the CAIP-2 ids match what each advertises on /supported.

Two negative runs: breaking the autoContracts skip fails the scaffolding
assertion, and breaking the plopfile guard fails two more. Both restore
green. A third negative probe passed and was discarded rather than
reported — it targeted the finalContractFramework fallback, which -t
makes unreachable.

Known gap, stated rather than papered over: the tests exercise the
non-interactive path, since -t forces it. The three when() clauses on
the interactive prompts are covered by reading, not by a test.

Build, lint and 7/7 tests pass. docs/templates/x402.mdx is added and
listed in docs.json — worth noting the previous template, ai-chat, has
no docs page.


ghozzza:
Pushed a fix and one finding that is not mine to fix here.

**Fixed (4758d7d):** `apps/api` declared a `lint` script but shipped no ESLint config, so `turbo lint` at the root failed for every generated x402 project. Added `.eslintrc.json` and the two `typescript-eslint` devDependencies, plus an assertion so the script and the config cannot drift apart again.

Found by running the generated project rather than reading it. The same pass also confirmed `tsc` emits correctly and `node dist/index.js` serves a real 402, so the ESM output and the `.js` import specifiers work under plain node, not just under `tsx`.

**Pre-existing, left alone:** `web:lint` fails on a freshly generated project on current `main` as well. Control run, no x402 involved:

```
$ celo-composer create demo-basic -t basic
$ pnpm lint
web:lint: ⚠ If you set up ESLint yourself, we recommend adding the Next.js ESLint plugin.
Failed: web#lint
```

`apps/web` has no `.eslintrc` and `next lint` exits 1 rather than scaffolding one non-interactively. It affects every template, so it predates this branch and fixing it here would mix two changes. Happy to open a separate issue if useful.

**Still not covered, stated plainly:** the buyer script has not been run end to end — that needs a funded Sepolia wallet, and what is verified so far is the 402 and its payment requirements, not a completed settlement. The three `when()` clauses on the interactive prompts are also covered by reading rather than by a test, since `-t` forces the non-interactive path.

ghozzza:
## End-to-end settlement, run against the generated project

The buyer script has now been run for real on Celo Sepolia. Everything below is read back from chain, not from the response body.

**tx** [`0x8a6a4deb6dc02c8b8535a773bc057e476577725780d7e0af80baf6321d8b5e2f`](https://celo-sepolia.blockscout.com/tx/0x8a6a4deb6dc02c8b8535a773bc057e476577725780d7e0af80baf6321d8b5e2f) — status `success`, block 33010775, gasUsed 102894

| | before | after |
|---|---|---|
| buyer `0x48a8eF42…Af31` USDC | 20 | 19.99 |
| payTo `0x10D400D1…Fff88` USDC | 0 | 0.01 |
| **buyer CELO** | **0** | **0** |

Receipt:

```
tx.from (pays gas) : 0x0d74d5cefd2e7f24e623330ebe3d8d4cb45ffb48   <- facilitator signer
tx.to              : 0x01c5c0122039549ad1493b8220cabedd739bc44e   <- USDC
logs               : 2
  AuthorizationUsed   authorizer=0x48a8ef42...af31
  Transfer            0x48a8ef42...af31 -> 0x10d400d1...fff88   0.01
```

Two things this pins down that reasoning could not:

**The buyer never held any CELO, before or after,** and the gas was paid by the facilitator signer rather than the payer. That is the claim the template is built around, and it is now on chain rather than in a paragraph.

**`AuthorizationUsed` is in the receipt**, so the EIP-3009 path executed against the token itself — the signature was built from `name="USDC"` / `version="2"`, the two values that cannot be read back from the contract. Reconstructing `DOMAIN_SEPARATOR` beforehand said those were right; this proves the whole signing path accepts them.

### What the run found before it worked

The first attempt failed with `401 Missing X-API-Key` from `/settle`, and that is worth recording because of the shape of it. `/supported` and `/verify` both answer **without** a key, so the server boots clean and returns a completely correct 402 — the failure only appears at settlement, after the buyer has signed. I had checked those two endpoints and concluded the key was optional on testnet, which was wrong.

Fixed in d698a49: `.env.template`, `x402.ts`, `X402_SETUP.md` and the docs page now say the key is required, and the seller warns at boot instead of letting it surface at settlement.

Also fixed earlier in this branch, both found the same way — by running the generated project rather than reading it:

- `apps/api` had a `lint` script and no ESLint config, so `turbo lint` failed for every generated project (4758d7d)
- nothing loaded `.env` at all, so the documented setup produced a free endpoint while printing "Set it in .env to charge" (ea92d00)

None of build, typecheck, the 402 probe or the test suite touches any of the three, because each of them supplies its own environment.

ghozzza:
## Correcting myself on the API key

I said above there was "no documented way for an outside developer to get one". That was wrong — I had not looked. There is a public, self-service path, and the facilitator documents it itself at [x402.celo.org/SKILL.md](https://x402.celo.org/SKILL.md):

1. Open [https://x402.celo.org](https://x402.celo.org), connect an EVM wallet
2. Click **Create API key**, sign the message — no gas, no transaction
3. Copy the key (`x402_…`), shown once

**New accounts start with free credits on both mainnet and testnet**, so this template runs end to end without anyone buying anything. That materially changes the template's story: a developer who picks `x402` out of the `create` menu can reach a real settlement without talking to a human.

Written into `.env.template`, `X402_SETUP.md` and the docs page in 9129d9d, along with two things that were missing and matter more than the steps:

**The key is server-side only, and the buyer needs none.** Worth stating explicitly in a scaffold that ships both halves in one repo, so nobody wires it into the client.

**Running out of credits fails at the same point as a missing key, but returns `402` instead of `401`** — which reads like a refused payment rather than an account balance. Documented as its own failure mode.

The test now asserts the setup guide names the dashboard, so the instructions cannot regress to "ask someone".

ghozzza:
## Closing the two gaps I flagged

Both were listed as "covered by reading, not by a test". Both are now run.

**The interactive prompt path.** `-t` forces the non-interactive branch, so the three `when()` clauses could not be reached from the test suite. Drove the real prompts through a PTY instead. Selecting **x402 Paid API** goes straight to `Install dependencies?` — the wallet provider and smart contract framework prompts are both skipped, which is what the three clauses are for:

```
? Which template would you like to use? x402 Paid API
? Install dependencies? No
```

The project it produced matches the flagged one exactly: `apps/api` present, `apps/contracts` absent, no wallet components in `apps/web`.

**Root `turbo build`.** 2/2 successful — `web` builds on Next 14 and `api` emits to `dist/`, with `dist/**` already in the pipeline's build outputs, so no `turbo.json` change was needed.

Remaining untested, and I would rather say it than imply otherwise: `X402_NETWORK=mainnet`. The constants switch together in one object and the testnet path is proven end to end, but the mainnet branch has not been executed and I am not going to move real USDC to prove a scaffold.

templates/base/package.json.hbs
```diff
@@ -13,8 +13,16 @@
     "contracts:test": "turbo run test --filter=hardhat",
     "contracts:deploy": "turbo run deploy --filter=hardhat",
     "contracts:deploy:celo-sepolia": "turbo run deploy:celo-sepolia --filter=hardhat",
-    "contracts:deploy:celo": "turbo run deploy:celo --filter=hardhat"{{/if}}
-  },
+    "contracts:deploy:celo": "turbo run deploy:celo --filter=hardhat"{{/if}}{{#if (eq templateType "x402")}},
+    "api:dev": "turbo run dev --filter=api",
+    "api:buy": "turbo run buy --filter=api"{{/if}}
```
GigaHierz, line 23: **Blocking:** this script cannot run. `turbo run buy` requires a `buy` entry in `turbo.json`'s `pipeline`, and `templates/base/turbo.json.hbs` isn't touched by this PR.

Generated a project from this branch and ran it:

```
$ pnpm api:buy
  x missing tasks in project

Error:   x could not find task `buy` in project
```

Not a missing-`node_modules` artifact — `turbo run type-check --filter=api` on the same tree gets all the way to executing the package script and fails on `Cannot find module 'viem'`. So `buy` is rejected at graph construction time.

`api:dev` is fine, because `dev` is already in the pipeline. Add the matching task in `templates/base/turbo.json.hbs`, guarded the same way these scripts are:

```hbs
    "clean": {
      "cache": false
    }{{#if (eq templateType "x402")}},
    "buy": {
      "cache": false
    }{{/if}}
```

Every other `turbo run X` script in this file — `compile`, `test`, `deploy`, `deploy:celo-sepolia`, `deploy:celo` — has a matching pipeline entry. This is the only one that doesn't.

src/generators/template-wiring.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    expect(x402).toContain('process.env.X402_NETWORK === "mainnet"');
+    // Without this, .env is never read and the documented setup silently
+    // produces a free endpoint. Node's own --env-file needs 20.6; base allows 18.
+    expect(x402).toContain('import "dotenv/config"');
+  });
+
+  it("declares the x402 scripts and the paywall peer exemption on the root package", () => {
+    const pkg = fs.readJsonSync(path.join(projectPath, "package.json"));
+    expect(pkg.scripts["api:dev"]).toBeDefined();
+    expect(pkg.scripts["api:buy"]).toBeDefined();
```
GigaHierz, line 108: This is why the suite is green despite `pnpm api:buy` being broken: it asserts the script is *defined*, not that it's runnable. `turbo run buy` fails at graph construction because `buy` isn't in `turbo.json`'s pipeline (see the comment on `templates/base/package.json.hbs`).

Given the argument this PR makes — that a template type has to be threaded through twelve places and a missed one fails silently — this is worth closing rather than leaving. `turbo.json` is a thirteenth place, and the test that exists to catch exactly this class of bug currently certifies past it.

Cheapest version, no subprocess needed:

```ts
const turbo = fs.readJsonSync(path.join(projectPath, "turbo.json"));
expect(turbo.pipeline.buy).toBeDefined();
```

templates/x402/apps/api/src/x402.ts.hbs
```diff
@@ -0,0 +1,76 @@
...
+/** CAIP-2 chain id, e.g. `eip155:11142220`. The `@x402` packages key schemes by this. */
+export const CAIP2: `eip155:${string}` = CONFIG.caip2;
+
+/**
+ * Price in RAW token units, not decimals. USDC has 6, so "10000" is 0.01 USDC.
+ * Getting this wrong by 10^6 is the single most common mistake when wiring x402.
+ */
+export const PRICE = process.env.X402_PRICE ?? "10000";
+
+/** Where settled payments land. Falls back to the buyer's own address in dev. */
```
GigaHierz: Stale comment — there's no such fallback. `PAY_TO` is just the env var, and unset means `/paid` answers for free, which `index.ts.hbs` warns about at boot and the `.env` template documents correctly. This line is the only place that disagrees, and "falls back to the buyer's own address" would be a meaningfully different (and slightly alarming) behaviour if someone believed it.

```suggestion
/** Where settled payments land. Unset means /paid answers for free. */
```

templates/x402/apps/api/src/index.ts.hbs
```diff
@@ -0,0 +1,90 @@
...
+              scheme: "exact",
+              network: CAIP2,
+              payTo: getAddress(PAY_TO),
+              price: {
+                amount: PRICE,
+                asset: CONFIG.usdc,
+                // The EIP-712 domain the token itself signs with. These two values
+                // cannot be read from the contract, so they are stated here — and if
+                // they are wrong the token rejects the signature at settlement time.
+                extra: { name: "USDC", version: "2" },
```
GigaHierz: These two values are hardcoded across both networks while everything else network-specific — facilitator, chain id, USDC address, explorer — is switched per network in the `X402` record in `x402.ts.hbs`.

The description says the Sepolia leg was verified live. The mainnet domain is therefore asserted rather than checked, and it's the one asymmetry in an otherwise per-network config. If mainnet USDC's EIP-712 `name`/`version` differ, users hit precisely the failure your own docs single out: rejected at settlement, after the buyer has signed, with everything upstream looking correct.

Suggest moving them into `X402` next to `usdc`:

```ts
testnet: {
  ...
  usdc: "0x01C5C0122039549AD1493B8220cABEdD739BC44E" as Address,
  domain: { name: "USDC", version: "2" },
},
```

Two benefits: the address and the domain it belongs to become impossible to desync, and the mainnet value becomes a reviewable line rather than an implication.

src/commands/create.ts
```diff
@@ -64,15 +64,17 @@ export async function createCommand(
             const autoWallet = options.walletProvider
               ? options.walletProvider
               : autoTemplateType === "farcaster-miniapp" ||
-                autoTemplateType === "ai-chat"
+                autoTemplateType === "ai-chat" ||
+                autoTemplateType === "x402"
```
GigaHierz: This defaults `walletProvider` to `"none"` for x402, but an explicit `--wallet-provider` still wins — while the new plopfile guards skip the rainbowkit and thirdweb component actions for x402 *unconditionally*. The two disagree, and the result is a scaffold that doesn't compile:

```
$ ... create x402-rk -t x402 --wallet-provider rainbowkit --skip-install
$ ls x402-rk/apps/web/src/components/
navbar.tsx  ui
$ grep -n connect-button x402-rk/apps/web/src/components/navbar.tsx
14:import { ConnectButton } from "@/components/connect-button"
```

The navbar's `{{#if (or (eq walletProvider "rainbowkit") ...)}}` guard fires, so it imports a component nothing writes.

Farcaster and Minipay have the same plopfile guards but ship their own `connect-button.tsx`, so they're safe. x402 ships none, which makes it the only combination with no substitute.

Low severity — few people will pass that flag — but it fails silently instead of erroring, which is the annoying kind. Cheapest fix is to *force* rather than default, so the flag can't contradict the guards:

```ts
const autoWallet =
  autoTemplateType === "farcaster-miniapp" ||
  autoTemplateType === "ai-chat" ||
  autoTemplateType === "x402"
    ? "none"
    : options.walletProvider ?? "rainbowkit";
```

Same reasoning applies to `--contracts` on the lines below.

templates/x402/apps/api/tsconfig.json.hbs
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
+    "lib": ["ES2022"],
+    "module": "ESNext",
+    "moduleResolution": "bundler",
```
GigaHierz: Nit: nothing bundles this — it's run by `tsx` in dev and `node dist/index.js` after `tsc`. `"bundler"` happens to work only because every relative import in the template already carries an explicit `.js` (`./x402.js`), but it *permits* extensionless imports rather than rejecting them, so the first person to add `import { x } from "./helpers"` gets a clean type-check and a runtime `ERR_MODULE_NOT_FOUND`.

`"NodeNext"` (with `"module": "NodeNext"`) describes what actually happens and enforces the extension you're already writing.

docs/templates/x402.mdx
```diff
@@ -0,0 +1,93 @@
...
+of the payment, and **the buyer needs no key at all**.
+
+### Then fill in `.env`
+
+Set `X402_API_KEY` to that key, `X402_PAY_TO` to an address you control — until it is
+set, `/paid` answers for free — and `BUYER_PRIVATE_KEY` to a throwaway key holding a
+little testnet USDC.
+
+```bash
+pnpm api:dev    # seller on http://localhost:3001
```
GigaHierz, line 63: These are the two scripts that don't work today (`api:buy` more permanently — see the comment on `templates/base/package.json.hbs`), while `X402_SETUP.md` in the generated project documents `pnpm --filter api dev` / `pnpm --filter api buy`, which do.

Worth converging on one form once the turbo pipeline entry lands. The generated project is where someone will actually be sitting, so `X402_SETUP.md` is arguably the one to match.

GigaHierz (changes requested):
This is a genuinely good PR — carefully reasoned, honestly documented, and it does the unglamorous work of threading a new template type through every place that needs it. I'm requesting changes for one concrete bug that ships broken to every user of the template, plus a few smaller things. Details inline; summary here.

## Blocking

**`pnpm api:buy` cannot work — `buy` was never added to the turbo pipeline.**

The new root script is `"api:buy": "turbo run buy --filter=api"`, but `templates/base/turbo.json.hbs` isn't in this diff and has no `buy` task. Turbo refuses to run a task that isn't declared.

Reproduced against this branch — generated a real project and ran it:

```
$ pnpm api:buy
  x missing tasks in project

Error:   x could not find task `buy` in project
```

Control, to rule out the missing-`node_modules` warning as the cause — a task that *is* in the pipeline gets as far as executing the package script:

```
$ turbo run type-check --filter=api
api:type-check: src/x402.ts(13,30): error TS2307: Cannot find module 'viem'...
```

So `buy` is rejected at graph construction; `type-check` runs. `api:dev` is fine because `dev` already exists in the pipeline. Fix is one entry in `templates/base/turbo.json.hbs`, guarded like the scripts are:

```hbs
    "clean": {
      "cache": false
    }{{#if (eq templateType "x402")}},
    "buy": {
      "cache": false
    }{{/if}}
```

Every other `turbo run X` script in that file already has a matching entry — `compile`, `test`, `deploy`, `deploy:celo-sepolia`, `deploy:celo`. `buy` is the only one that doesn't.

Two knock-ons worth fixing in the same pass:

- `docs/templates/x402.mdx` tells people `pnpm api:dev` / `pnpm api:buy`, while `X402_SETUP.md` tells them `pnpm --filter api dev` / `pnpm --filter api buy`. Right now only the second pair works. Once the pipeline entry lands they'll agree, but they should say the same thing either way.
- The test asserts `pkg.scripts["api:buy"]` is *defined*, never that it's runnable — which is why the suite is green. Since the PR's own argument is that generating-and-asserting is what catches missed wiring, this is worth closing: assert the turbo task exists, or invoke the script.

## Also confirmed by generating projects

**`README.md` still lists the old template set.** `src/index.ts` and the docs site were updated; the README is the thirteenth place — which is rather the point the description makes.

- line 12 — "Basic Web App, Farcaster Miniapp, Minipay, or AI Chat templates"
- line 143 — ``| `-t, --template <type>` | Template type (`basic`, `farcaster-miniapp`, `minipay`, `ai-chat`) | ``

An `### x402 Paid API` section alongside the existing per-template sections would be consistent too.

**`-t x402 --wallet-provider rainbowkit` generates a project that can't compile.** Verified:

```
$ ... create x402-rk -t x402 --wallet-provider rainbowkit --skip-install
$ ls x402-rk/apps/web/src/components/
navbar.tsx  ui
$ grep -n connect-button x402-rk/apps/web/src/components/navbar.tsx
14:import { ConnectButton } from "@/components/connect-button"
```

The new plopfile guards skip the rainbowkit/thirdweb components for x402 unconditionally, but `create.ts` still honours an explicit `--wallet-provider`, so the navbar imports a file nothing writes. Farcaster and Minipay carry the same guards but ship a replacement component — x402 is the one combination with no substitute. Low severity, but it fails silently rather than erroring. See inline for the cheapest fix.

## Worth saying, since a review shouldn't be only findings

- The `import "dotenv/config"` placement, with the comment explaining *why* it has to be first, is the kind of thing that saves the next reader an hour of confusion.
- The `.env` template documenting the "`/supported` and `/verify` answer without a key, only `/settle` 401s — after the buyer has already signed" trap is the most valuable thing in this PR. That's a failure mode you only learn by hitting it.
- The `apps/api`-rather-than-a-route reasoning is sound. Pinning Next 16 for the whole base web app to accommodate one seller would have been the worse trade, and the write-up makes the case properly instead of asserting it.
- Tests that generate real projects and assert on the output beat tests that assert on source. One correction to the framing: `jest.config.cjs` was already configured with `testMatch: ['<rootDir>/src/**/*.test.ts']` and a `ts-jest` transform, so the harness was waiting for this rather than being established by it — which if anything strengthens the case for adding the file.
- Naming the three open scope questions and picking defaults out loud, rather than blocking on them, is the right call.

ghozzza:
All addressed at `500071b`. Two places where I ended up somewhere different from the suggestion, both with evidence below.

## Blocking — `pnpm api:buy`

Fixed exactly as suggested. Your repro now gets to the script:

```
api:buy: > tsx src/buyer.ts
api:buy: BUYER_PRIVATE_KEY is not set. Copy .env.template to .env first.
```

Which is the correct failure for a scaffold with no `.env` — the point is that it *reaches* the buyer instead of being rejected at graph construction.

**And the test.** You were right that it certified past the bug. It now asserts `turbo.pipeline.buy` and `turbo.pipeline.dev`, and I checked it actually bites: removing the task again turns the suite red rather than green.

## The wallet flag — your fix was necessary but not sufficient

Applying the suggested change to `autoWallet` alone, then generating, still reproduced:

```
$ ls x402rk/apps/web/src/components/
navbar.tsx  ui
$ grep -n connect-button x402rk/apps/web/src/components/navbar.tsx
14:import { ConnectButton } from "@/components/connect-button"
```

There is a **second** resolution further down, and it is the one that reaches plop:

```ts
const finalWalletProvider =
  options.walletProvider ||          // <- explicit flag wins again, here
  answers.walletProvider || ...
```

So the auto-mode branch computes `"none"` and then this line overrides it back. Forced at `finalWalletProvider` instead, which is the single authoritative point. Verified after: x402 and ai-chat with either `--wallet-provider` now emit no `connect-button` import at all.

Flagging it because the suggested diff looks complete in isolation — I only caught it by generating a project afterwards rather than reading the change.

## `--contracts` — I did **not** apply the same reasoning, and I think it would be a regression

The two guard sets are not symmetric. The wallet actions in `plopfile.ts` check the template type as well:

```
"Skipping RainbowKit - This template uses its own wallet components"
```

The contract actions check **only** `contractFramework`:

```ts
if (data.contractFramework !== "hardhat") { return "Skipping Hardhat - ..." }
```

So an explicit `--wallet-provider` genuinely cannot be honoured, while an explicit `--contracts` genuinely can. `-t farcaster-miniapp -c hardhat` scaffolds `apps/contracts` correctly today — verified, still does after this change — and the interactive prompt offers the contract question for farcaster (only `ai-chat` and `x402` return false). Forcing it to `none` would remove a working, offered combination.

Left as a default rather than a force, with the asymmetry written down next to it. Happy to change it if you read the guards differently.

## The EIP-712 domain

Moved into the per-network record as suggested. One correction to the comment that was there: it claimed the values *cannot* be read from the contract. They can —

| | `name()` | `version()` | `eip712Domain()` |
|---|---|---|---|
| mainnet `0xcebA93…118C` | `USDC` | `2` | reverts |
| sepolia `0x01C5C0…C44E` | `USDC` | `2` | reverts |

So the mainnet value is no longer asserted; it is what the contract answers. `eip712Domain()` reverting on both is the reason the original claim looked true, and it is now written down so the next person checks the right two getters.

## Command form — went the other way

You suggested converging on `X402_SETUP.md`'s `pnpm --filter api …`. I converged on the root scripts instead, and updated the four places that used the other form.

Reason: this repo already established that pattern. `contracts:compile`, `contracts:test`, `contracts:deploy*` are all root scripts, and the hardhat README documents `pnpm contracts:compile` rather than the filter form. `api:dev` / `api:buy` matching that is more consistent than making the x402 template the one place that documents the underlying invocation — and now that the pipeline entry exists, they work.

Easy to flip if you would rather the generated `X402_SETUP.md` speak in `--filter` terms.

## The rest

- **`tsconfig` → `NodeNext`** for both `module` and `moduleResolution`. Type-checks clean on a generated project, which it should, since every relative import already carries `.js`.
- **Stale `PAY_TO` comment** — took your suggestion verbatim.
- **README** — template list at line 12, the flag table, and an `### x402 Paid API` section alongside the others.
- **A fourteenth place.** `docs/cli-reference/create-command.mdx:26` also enumerated the old template set. Found by sweeping for the enumeration rather than fixing the one you named — which is rather the point the description was making.

Thank you for reproducing the turbo failure rather than describing it. The control run against `type-check` is what made it unambiguous that it was graph construction and not the missing `node_modules`.

ghozzza:
One merge-order note, since #442 landed after this PR was opened.

**#442 adds `.choices()` validation to `-t`, `--wallet-provider` and `-c`**, so an unknown value is rejected with the allowed list instead of silently producing a bare base project (#412). Its `-t` list is `basic, farcaster-miniapp, minipay, ai-chat` — **it does not include `x402`**, because on `main` that template does not exist yet and accepting the value would reproduce exactly the bug it fixes.

So whichever of these two lands second needs one line:

- **this PR first** → add `'x402'` to the `-t` choices in #442
- **#442 first** → add `'x402'` here, in the `.addOption(new Option('-t, --template <type>')...)` block

CI will not catch it either way: a missing choice is a rejected flag at runtime, not a failing build. I will make the edit on whichever side is second as soon as the other merges.

GigaHierz:
Re-reviewed after the recent pushes — the earlier blockers (eslint config, dotenv loading, API-key docs, `api:buy` turbo task, wallet-flag forcing, `-y` in the test fixture) are all verified fixed, and the payment flow itself checked out end-to-end (middleware verifies-then-settles server-side before releasing the body; facilitators advertise `eip155:42220`/`eip155:11142220`; both USDC contracts confirmed on-chain with `name()="USDC"`, `version()="2"`, matching the EIP-712 domain in `extra`). This is in good shape.

Remaining coordination + small items:

1. **Merge order/conflicts (now real):** current heads conflict with #440 (`src/commands/create.ts` auto-mode block) and #442 (`src/index.ts` `-t` option line). Recommended: land #440 and #442 first, then rebase this once — adding `'x402'` to #442's `.choices()` list and re-threading the x402 `when:` conditions onto #440's `cliTemplateType` normalization (the current `when:` clauses read `options.templateType`, which misses `-t`).
2. **Silent flag discard:** `create.ts` now forces `walletProvider = "none"` for farcaster/ai-chat/x402 — correct, but an explicit `--wallet-provider thirdweb -t x402` is ignored without a word. A one-line `console.warn` would save users confusion.
3. Nits: the rainbowkit/thirdweb skip messages say the template "uses its own wallet components", which is inaccurate for x402 (it ships none); the thirdweb *lib* action lacks the x402 guard the components action got (harmless today, asymmetric); the `"//"` comment key sits inside `pnpm.peerDependencyRules` (tolerated, nonstandard spot).

GigaHierz:
#440 and #442's queue position is resolved (#440 merged; #442 needs only a trivial rebase). Once #442 lands, please rebase: add 'x402' to the -t choices list and re-thread the x402 when: clauses onto cliTemplateType per the earlier note. The changes-requested review will be dismissed after that rebase.

GigaHierz:
Rebased onto main as a maintainer edit, resolving all five conflicts per the agreed plan: 'x402' added to #442's -t choices list, the when: clauses and auto-branch re-threaded onto #440's cliTemplateType normalization (keeping this branch's wallet-forcing logic and its rationale comment), the foundry + x402 script blocks united in package.json.hbs, both docs pages kept in docs.json, and tsconfig excludes united. Verified post-rebase: build passes, full test run is 22/22 (the main smoke suite + this branch's template-wiring tests), `-t bogus` still rejected, and an x402 scaffold generates with valid manifests, both workspaces, the api:dev/api:buy scripts, and the peerDependencyRules block. Ready for review dismissal + merge.

ghozzza:
Thanks for taking the rebase — the resolution matches what we agreed, and I have not pushed anything on top so what you verified is still what is there.

Two items from your 12 Aug review are not in it, and I did not want to invalidate a verified branch on the way to a merge, so they are **#453** (Backlog on the board): the missing `console.warn` when an explicit `--wallet-provider` is discarded, and the skip message reading "this template uses its own wallet components" for x402, which is the one template that ships none. I added the third nit you spotted while confirming it — the thirdweb *lib* action guards only `farcaster-miniapp` while the components action guards farcaster and x402. Latent rather than live, since the forcing means the wallet check catches it today.

The `"//"` key in `pnpm.peerDependencyRules` I left alone, since you called it tolerated.

Held #453's PR until this merges — two of its three files are in this diff, and the x402 wording describes a template that is not on main yet.

GigaHierz (approved):

PR #394, opened 2026-08-10
fix(templates): ship an ESLint config so `next lint` can run

### LINKED ISSUE #395: web:lint fails on every generated project — apps/web ships no ESLint config
### What happens

`pnpm lint` fails at the root of every generated project that has an `apps/web`.

```
$ npx @celo/celo-composer create demo -t basic
$ cd demo && pnpm install && pnpm lint

web:lint: ? How would you like to configure ESLint?
web:lint:  ❯ Strict (recommended)
web:lint:    Base
web:lint:    Cancel
web:lint:  ELIFECYCLE  Command failed with exit code 1.
Failed: web#lint
```

### Why

`apps/web` declares `"lint": "next lint"` but ships no ESLint config. With no config, `next lint` falls through to its interactive setup wizard. Under `turbo lint` — or in any CI — there is no TTY to answer it, so the task exits 1.

### Scope

Every template with an `apps/web` **except** `farcaster-miniapp`, which happens to ship its own `.eslintrc.json` and is therefore the only one whose lint runs. `ai-chat` is unaffected — it is a standalone template with no `apps/web`.

### Fix

Move the same config into the base template. `eslint-config-next` is already a devDependency there, so nothing new is added.

One wrinkle worth noting for review: the Farcaster action has to become `force: true`, otherwise the base file lands first and Farcaster silently loses its three rule overrides.

PR: #394

ghozzza:
`apps/web` has a `lint` script but no ESLint config, so `next lint`
falls through to its interactive setup wizard:

  ? How would you like to configure ESLint?
    Strict (recommended) / Base / Cancel

There is no TTY under `turbo lint` or in CI, so it cannot be answered
and the task exits 1. Every generated project fails `pnpm lint` at the
root, on every template that has an `apps/web`.

The repo already had the answer for exactly one template:
`templates/farcaster-miniapp` ships its own `.eslintrc.json` and is the
only one whose lint runs. This moves the same config into the base
template, where `eslint-config-next` is already a devDependency.

The Farcaster action now sets `force: true`. Without it the base file
lands first and Farcaster silently loses its three rule overrides —
caught by generating every template rather than only `basic`. Same
idiom the Minipay tailwind config already uses.

Verified per template after generating each one:

  basic              extends next/core-web-vitals, 0 rule overrides
  minipay            extends next/core-web-vitals, 0 rule overrides
  farcaster-miniapp  extends next/core-web-vitals, 3 rule overrides
  ai-chat            unaffected — standalone template, no apps/web

Scope note: this makes `next lint` *run*. It does not make it pass.
Once the wizard is out of the way it reports two real errors in
`navbar.tsx`, which are a separate bug with a separate cause — the
navbar imports `ConnectButton` while the component exports
`WalletConnectButton`. That is independently reproducible with
`tsc --noEmit` and has nothing to do with ESLint, so it is not fixed
here. Filed separately.

Closes #395.

### DIFF
--- a/src/plopfile.ts
+++ b/src/plopfile.ts
@@ -214,6 +214,9 @@ export default function (plop: NodePlopAPI): void {
         globOptions: {
           dot: true,
         },
+        // The base template now ships an .eslintrc.json of its own, so this has to
+        // overwrite it rather than silently lose the Farcaster rule overrides.
+        force: true,
         skip: (data: PlopData): string | false => {
           if (data.templateType !== "farcaster-miniapp") {
             return "Skipping Farcaster Miniapp config - different template type selected";
--- a/templates/base/apps/web/.eslintrc.json.hbs
+++ b/templates/base/apps/web/.eslintrc.json.hbs
@@ -1,0 +1,3 @@
+{
+  "extends": "next/core-web-vitals"
+}


src/plopfile.ts
```diff
@@ -214,6 +214,9 @@ export default function (plop: NodePlopAPI): void {
         globOptions: {
           dot: true,
         },
+        // The base template now ships an .eslintrc.json of its own, so this has to
+        // overwrite it rather than silently lose the Farcaster rule overrides.
+        force: true,
```
GigaHierz, line 219: Confirmed this is necessary rather than defensive: this action is declared after the base `addMany`, and `plop-generator.ts` builds nodePlop with `force: false`. Without the flag the base `.eslintrc.json` lands first and Farcaster silently drops its three rule overrides — no error, just quieter linting. Same pattern as the Minipay tailwind override above.

GigaHierz (approved):
Approving. I checked the two things that could quietly make this a no-op, and both are fine.

**The dotfile actually gets picked up.** A `.hbs` template starting with a dot silently not matching the glob is the usual way this kind of change fails. It doesn't here — the base `addMany` in `src/plopfile.ts` already sets `globOptions: { dot: true }`, so `.eslintrc.json.hbs` is emitted.

**`force: true` is required, not defensive.** Worth spelling out why, since it looks like a belt-and-braces addition: the Farcaster action is declared *after* the base `addMany` in the actions array, and `plop-generator.ts` constructs nodePlop with `force: false`. Without the flag the base file wins and Farcaster silently loses its three rule overrides — a regression that produces no error, just weaker linting. Good catch, and it matches the existing idiom for the Minipay tailwind override.

**"ai-chat unaffected" is accurate for a reason not stated in the description:** `ai-chat` never reaches plop at all. `plop-generator.ts` has a raw-copy fast path that returns before the generator runs, so the whole actions array is bypassed. The conclusion is right; the mechanism is stronger than "no `apps/web`".

**No new dependency** — `eslint-config-next` is already in `templates/base/apps/web/package.json.hbs` devDependencies, so this is purely wiring up something that was already paid for.

I also appreciate the scope discipline in the description: making `next lint` *run* and explicitly not making it *pass*, with the remaining `navbar.tsx` errors split out into #397 rather than bundled in. Two small reviewable PRs beats one that mixes causes.

One forward-looking note, not blocking: this is the eslint 8 legacy `.eslintrc` format, which is correct today given the pinned `next ^14.0.0` / `eslint-config-next ^14.0.0`. Whenever the base template moves to Next 15/16, this file and the Farcaster one both need to become flat config (`eslint.config.mjs`), and `next lint` itself is deprecated in Next 16 in favour of the ESLint CLI. Worth a note on whatever tracks the Next upgrade — not something to solve here.

GigaHierz (approved):

PR #397, opened 2026-08-10
fix(templates): navbar references a component that is never defined

### LINKED ISSUE #396: navbar renders <WalletConnectButton /> but imports ConnectButton — undefined component in every generated app
### What happens

`turbo type-check` is red on the default scaffold, for every wallet provider.

```
$ npx @celo/celo-composer create demo -t basic
$ cd demo && pnpm install && pnpm --filter web type-check

navbar.tsx(14,10): error TS2459: Module '"@/components/connect-button"' declares 'ConnectButton' locally, but it is not exported.
navbar.tsx(59,20): error TS2552: Cannot find name 'WalletConnectButton'. Did you mean 'ConnectButton'?
navbar.tsx(94,14): error TS2552: Cannot find name 'WalletConnectButton'. Did you mean 'ConnectButton'?
```

`navbar.tsx` imports `ConnectButton` and then renders `<WalletConnectButton />` three times. The import is unused and the rendered component is undefined.

### Why

A naming inconsistency rather than a typo. Three of the four connect-button components export `WalletConnectButton`:

| component | exports |
| --- | --- |
| `wallets/rainbowkit` | `WalletConnectButton` |
| `wallets/thirdweb` | `WalletConnectButton` |
| `farcaster-miniapp` | `WalletConnectButton` |
| `minipay` | `ConnectButton` |

The navbar was written against the odd one out while rendering the majority name, so it is wrong in both directions depending on the template: with rainbowkit or thirdweb the import is wrong, and with Minipay the render is wrong. Either way the component does not exist at render time.

### Suggested fix

Normalise Minipay onto `WalletConnectButton` and point the navbar import at it. The three render sites are already correct.

PR: #397

ghozzza:
`apps/web/src/components/navbar.tsx` imports `ConnectButton` and then
renders `<WalletConnectButton />` three times. The two names are not the
same symbol, so the generated app has an unused import and an undefined
component.

`tsc --noEmit` on a default `create demo -t basic` from main:

  navbar.tsx(14,10): error TS2459: Module '"@/components/connect-button"'
    declares 'ConnectButton' locally, but it is not exported.
  navbar.tsx(59,20): error TS2552: Cannot find name 'WalletConnectButton'.
  navbar.tsx(94,14): error TS2552: Cannot find name 'WalletConnectButton'.

So `turbo type-check` is red on the default scaffold, for every wallet
provider.

The underlying cause is a naming inconsistency rather than a typo.
Three of the four connect-button components export
`WalletConnectButton` — rainbowkit, thirdweb, farcaster-miniapp — and
only Minipay exports `ConnectButton`. The navbar was written against
the odd one out while rendering the majority name.

This normalises Minipay onto `WalletConnectButton` and points the
navbar import at it. The three render sites are already correct and are
untouched. Minipay's `wallet-provider.tsx` imported the old name and
never used it, so that import is dropped rather than renamed to
something equally unused.

Verified by generating each combination and type-checking:

  basic + rainbowkit    import matches export, tsc clean
  basic + thirdweb      import matches export, tsc clean
  minipay               import matches export, tsc clean

Unrelated and not addressed here: `next build` on the same scaffold
fails separately with an unmet `@x402/evm` peer reached through
wagmi -> rainbowkit. That reproduces on main without this change.

Closes #396.

### DIFF
--- a/templates/minipay/components/connect-button.tsx.hbs
+++ b/templates/minipay/components/connect-button.tsx.hbs
@@ -4,7 +4,7 @@
 import { ConnectButton as RainbowKitConnectButton } from "@rainbow-me/rainbowkit";
 import { useEffect, useState } from "react";
 
-export function ConnectButton() {
+export function WalletConnectButton() {
   const [isMinipay, setIsMinipay] = useState(false);
 
   useEffect(() => {
--- a/templates/base/apps/web/src/components/navbar.tsx.hbs
+++ b/templates/base/apps/web/src/components/navbar.tsx.hbs
@@ -12,7 +12,7 @@ import {
   SheetTrigger,
 } from "@/components/ui/sheet"
 {{#if (or (eq walletProvider "rainbowkit") (eq walletProvider "thirdweb"))}}
-import { ConnectButton } from "@/components/connect-button"
+import { WalletConnectButton } from "@/components/connect-button"
 {{/if}}
 
 const navLinks = [
--- a/templates/minipay/components/wallet-provider.tsx.hbs
+++ b/templates/minipay/components/wallet-provider.tsx.hbs
@@ -9,7 +9,6 @@ import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { useEffect, useState } from "react";
 import { WagmiProvider, createConfig, http, useConnect } from "wagmi";
 import { celo, celoSepolia } from "wagmi/chains";
-import { ConnectButton } from "./connect-button";
 
 const connectors = connectorsForWallets(
   [


templates/minipay/components/connect-button.tsx.hbs
```diff
@@ -4,7 +4,7 @@
 import { ConnectButton as RainbowKitConnectButton } from "@rainbow-me/rainbowkit";
 import { useEffect, useState } from "react";
 
-export function ConnectButton() {
+export function WalletConnectButton() {
```
GigaHierz, line 7: This is the load-bearing line. Minipay defaults to `walletProvider = "rainbowkit"` in `create.ts`, so the navbar's guarded import block fires for Minipay too — and plopfile writes *this* file to `apps/web/src/components/connect-button.tsx`. So `@/components/connect-button` resolves here for Minipay scaffolds, and without this rename Minipay would still be broken even after the navbar import was fixed.

GigaHierz (approved):
Approving. I re-derived the diagnosis independently rather than taking the description on trust, and it holds up.

**The naming inconsistency is real and this is the right direction.** Three of the four connect-button components already export `WalletConnectButton` — `templates/wallets/rainbowkit`, `templates/wallets/thirdweb`, `templates/farcaster-miniapp`. Minipay was the only one exporting `ConnectButton`. Normalising the outlier onto the majority name is better than changing the three render sites to match the odd one out.

**The fix is complete.** All three `<WalletConnectButton />` render sites and the import sit inside the *same* `{{#if (or (eq walletProvider "rainbowkit") (eq walletProvider "thirdweb"))}}` guard, so import and usage stay in lockstep for every combination — there's no branch where one is emitted without the other.

**One thing worth stating explicitly, because it's the non-obvious part of this PR:** the Minipay rename isn't cosmetic tidying, it's load-bearing. Minipay gets `walletProvider = "rainbowkit"` from `src/commands/create.ts`, so the guarded import block *does* fire for Minipay — and `src/plopfile.ts` writes Minipay's own `connect-button.tsx` into `apps/web/src/components/`. So `@/components/connect-button` resolves to the Minipay file for those scaffolds. Renaming the export is what actually fixes Minipay; leaving it would have fixed only rainbowkit and thirdweb.

**The dropped import is genuinely dead.** `templates/minipay/components/wallet-provider.tsx.hbs` had exactly one reference to `ConnectButton` — the import line. No JSX usage. Removing rather than renaming is correct.

**No collateral damage:** `templates/ai/chat-template/components/wallet-connect.tsx` also mentions `ConnectButton`, but that's thirdweb's own export from `thirdweb/react`, unrelated to this symbol, and correctly untouched.

GigaHierz (approved):

PR #432, opened 2026-08-11
fix(templates): key MiniPay token addresses by chain, and export them

### LINKED ISSUE #408: MiniPay UserBalance hardcodes mainnet token addresses with no chain guard — silent empty reads on Celo Sepolia
## What happens

The MiniPay template's `UserBalance` component hardcodes **mainnet** token addresses with no chain guard, while the wagmi config it ships includes Celo Sepolia. On testnet, the component silently reads balances from contracts that don't exist there.

## Details

`templates/minipay/components/user-balance.tsx.hbs:6-8` hardcodes cUSD `0x765d...282a`, USDC `0xcebA...118C`, USDT `0x4806...3D5e` — all correct for Celo mainnet (42220) and matching the table in `docs/templates/minipay.mdx:119-125` — but the reads are issued regardless of the connected chain, and the wallet provider config includes `celoSepolia`.

Result on Celo Sepolia: `useReadContract` balance calls target nonexistent contracts and the UI shows empty/zero balances with no hint why. This is exactly the situation a new developer testing against the faucet will hit first.

## Suggested fix

Key the token addresses by chain id (`42220` / `11142220`) and either resolve per connected chain or clearly gate the component to mainnet. Exporting the address constants would also let users reuse them (see `docs/templates/minipay.mdx:115`, which currently tells users to reuse a constant that is module-local and unexported).

ghozzza:
Closes #408.

## The bug

`UserBalance` hardcoded three **mainnet** addresses and issued the reads regardless of the connected chain, while the wallet provider config ships `[celo, celoSepolia]`. On Sepolia those `useBalance` calls target contracts that do not exist there — which returns **zero, not an error**. The UI shows an empty wallet and gives no reason, to precisely the developer who just got faucet funds.

## The fix

Addresses are now keyed by chain id and **exported**:

```ts
export const TOKENS_BY_CHAIN: Record<number, { symbol: string; address: `0x${string}` }[]> = {
  [celo.id]:        [ USDm, USDC, USDT ],
  [celoSepolia.id]: [ USDC ],
};
```

Exporting is not cosmetic — `docs/templates/minipay.mdx:115` already tells readers *"The `UserBalance` component in the template already imports `USDT_ADDRESS` — use the same constant"*, and the constant was module-local and unexported, so that instruction could not be followed.

A chain with no entry now renders the native balance plus a line saying the app has no token list for it, rather than a silent column of zeros.

## A finding while verifying: cUSD now reports USDm

I read the symbols and decimals off the contracts instead of copying the doc table, against `forno.celo.org`:

| address | `name()` | `symbol()` | `decimals()` |
|---|---|---|---|
| `0x765DE816845861e75A25fCA122bb6898B8B1282a` | **Mento Dollar** | **USDm** | 18 |
| `0xcebA9300f2b948710d2653dD7B07f33A8B32118C` | — | USDC | 6 |
| `0x48065fbBE25f71C9282ddf5e1cD6D6A887483D5e` | Tether USD | USD₮ | 6 |

The template labelled the first one **cUSD**. It is the same token, renamed on chain, so the label now follows `symbol()` with a comment noting the former name. **The token table in `docs/templates/minipay.mdx:119-125` still says cUSD and needs the same correction** — that is a docs change, so I have left it for the docs tier rather than widening this PR.

The Sepolia USDC entry is verified the same way against `forno.celo-sepolia.celo-testnet.org` (chain id 11142220, `symbol()` = USDC, 6 decimals) rather than assumed. Sepolia has no equivalent for the other two, so the list is deliberately short instead of padded with addresses I could not confirm.

Also canonicalised the checksum on the first address — it was written all-lowercase in the code and checksummed in the comment two lines below.

## Verified

`create demo -t minipay -c none`, then `pnpm type-check`: `user-balance.tsx` is clean. The only remaining errors in that scaffold are

```
src/components/navbar.tsx(60,22): error TS2552: Cannot find name 'WalletConnectButton'.
src/components/navbar.tsx(96,14): error TS2552: Cannot find name 'WalletConnectButton'.
```

which are #396, fixed in #397 and awaiting review.

### DIFF
--- a/docs/templates/minipay.mdx
+++ b/docs/templates/minipay.mdx
@@ -112,7 +112,17 @@ async function sendUSDT(to: `0x${string}`, amount: string) {
 }
 ```
 
-The `UserBalance` component in the template already imports `USDT_ADDRESS` — use the same constant to keep the address consistent across your app.
+The `UserBalance` component exports `TOKENS_BY_CHAIN` — import that instead of writing an address down again, and it keeps working when the user is on Celo Sepolia rather than mainnet:
+
+```ts
+import { TOKENS_BY_CHAIN } from "@/components/user-balance";
+import { useAccount } from "wagmi";
+
+const { chainId } = useAccount();
+const usdt = TOKENS_BY_CHAIN[chainId!]?.find((t) => t.symbol === "USD₮")?.address;
+```
+
+The symbol is `USD₮` with U+20AE, which is what both Tether contracts return from `symbol()` — not the ASCII `USDT`.
 
 ## Token Reference
 
--- a/templates/minipay/components/user-balance.tsx.hbs
+++ b/templates/minipay/components/user-balance.tsx.hbs
@@ -16,9 +16,16 @@ import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
  * Exported so the rest of your app can reuse the same addresses instead of
  * copying them.
  *
- * Symbols and decimals below were read from the contracts themselves rather
- * than copied from a table. Note that `0x765DE8…282a` now reports **USDm**
- * ("Mento Dollar") — it is the token long known as cUSD, renamed on chain.
+ * Every symbol and decimal count below was read from the contract by calling
+ * `symbol()`, `name()` and `decimals()` on the chain named, not copied from a
+ * table. Two of them surprise people:
+ *
+ *   - `0x765DE8…282a` reports **USDm** ("Mento Dollar"). It is the token long
+ *     known as cUSD, renamed on chain. MiniPay's own UI still says cUSD.
+ *   - Both Tether contracts report **USD₮**, with U+20AE, not the ASCII "USDT".
+ *
+ * Mainnet and Sepolia carry the same three tokens, so a balance panel that
+ * works against the faucet shows the same rows in production.
  */
 export const TOKENS_BY_CHAIN: Record<
   number,
@@ -27,10 +34,12 @@ export const TOKENS_BY_CHAIN: Record<
   [celo.id]: [
     { symbol: "USDm", address: "0x765DE816845861e75A25fCA122bb6898B8B1282a" }, // 18 decimals, formerly cUSD
     { symbol: "USDC", address: "0xcebA9300f2b948710d2653dD7B07f33A8B32118C" }, // 6 decimals
-    { symbol: "USDT", address: "0x48065fbBE25f71C9282ddf5e1cD6D6A887483D5e" }, // 6 decimals
+    { symbol: "USD₮", address: "0x48065fbBE25f71C9282ddf5e1cD6D6A887483D5e" }, // 6 decimals
   ],
   [celoSepolia.id]: [
+    { symbol: "USDm", address: "0xdE9e4C3ce781b4bA68120d6261cbad65ce0aB00b" }, // 18 decimals
     { symbol: "USDC", address: "0x01C5C0122039549AD1493B8220cABEdD739BC44E" }, // 6 decimals
+    { symbol: "USD₮", address: "0xd077A400968890Eacc75cdc901F0356c943e4fDb" }, // 6 decimals
   ],
 };
 
--- a/templates/minipay/components/user-balance.tsx.hbs
+++ b/templates/minipay/components/user-balance.tsx.hbs
@@ -1,11 +1,38 @@
 "use client";
 
 import { useAccount, useBalance } from "wagmi";
+import { celo, celoSepolia } from "wagmi/chains";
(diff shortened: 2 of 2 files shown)


GigaHierz (changes requested):
The core fix is right and the Sepolia USDC address checks out on-chain (`symbol()` = "USDC", 6 decimals). Two things block this:

1. **The "Sepolia has no equivalent for the other two" rationale is incorrect.** Both tokens have official Celo Sepolia deployments, verified on-chain via `forno.celo-sepolia.celo-testnet.org`:
   - USDm: `0xdE9e4C3ce781b4bA68120d6261cbad65ce0aB00b` (`symbol()` = "USDm", 18 decimals)
   - USD₮: `0xd077A400968890Eacc75cdc901F0356c943e4fDb` (`symbol()` = "USD₮", 6 decimals)

   Please add all three to the Sepolia map for parity with mainnet — otherwise a faucet user's USDm balance still silently shows nothing, which is the exact UX this PR sets out to fix.

2. **This PR invalidates a doc line cited by the issue it closes.** `docs/templates/minipay.mdx:115` says "The `UserBalance` component in the template already imports `USDT_ADDRESS` — use the same constant." Before this PR that constant existed (unexported); after it, it doesn't exist at all. Issue #408 explicitly cites that line, so the one-line doc update (point it at the new exported structure) belongs in this PR rather than a follow-up.

Also worth a maintainer's explicit sign-off: relabeling cUSD → USDm in the UI is factually correct on-chain (the mainnet contract now returns `symbol() = "USDm"`), but MiniPay's own UI still says cUSD — a deliberate product decision either way.

ghozzza:
All at `52f1de4`. Your addresses check out, and re-running the same check on the mainnet entries found the map was already wrong about one of them.

## 1. The Sepolia rationale — you were right, and I verified before trusting it

Called `symbol()`, `name()` and `decimals()` directly against `forno.celo-sepolia.celo-testnet.org` (chain id **11142220**), plus `eth_getCode` so a wrong address cannot pass as an empty response:

| address | `symbol()` | `name()` | `decimals()` | code |
|---|---|---|---|---|
| `0xdE9e4C3ce781b4bA68120d6261cbad65ce0aB00b` | `USDm` | Mento Dollar | **18** | 1,765 bytes |
| `0xd077A400968890Eacc75cdc901F0356c943e4fDb` | `USD₮` | Tether USD | **6** | 2,227 bytes |
| `0x01C5C0122039549AD1493B8220cABEdD739BC44E` | `USDC` | USDC | 6 | 1,798 bytes |

All three are in the Sepolia map now, so it has parity with mainnet and a faucet user's USDm balance stops silently reading as nothing.

## 2. What the same check found on mainnet

The file's comment claimed *"symbols and decimals below were read from the contracts themselves rather than copied from a table."* That was not true of one entry:

```
0x48065fbBE25f71C9282ddf5e1cD6D6A887483D5e   symbol() = USD₮   (U+20AE, not ASCII "USDT")
```

The map said `"USDT"`. It is now `USD₮` on both chains, and the comment says why so the next person does not "correct" it back. This is the kind of thing the claim was supposed to prevent, and it slipped through because the claim was written before the check was run on every row.

## 3. `docs/templates/minipay.mdx` — done here, as you said

That line pointed at `USDT_ADDRESS`, which this PR deletes. It now points at the exported `TOKENS_BY_CHAIN` with a snippet that resolves through `chainId`, so the example keeps working on Sepolia instead of hardcoding a mainnet address — which is the whole reason the constant went away. It also notes the `USD₮` spelling, since anyone matching on `"USDT"` gets `undefined`.

Test-merged **#444**, which edits the token table further down the same file: **merges clean**.

## 4. The relabel — your call, and I have not made it

Deliberately left for you. On chain `0x765DE8…282a` returns `symbol() = "USDm"` and `name() = "Mento Dollar"`, so the relabel is factually right and the generated app now matches what a block explorer shows. But **MiniPay's own UI still says cUSD**, so a user reading "USDm" in a template app and "cUSD" in the wallet has to work out they are the same token.

Both readings are defensible and it is a product decision rather than a code one. The comment in the file states the situation without arguing for either, so flipping it back is a one-line change if you decide the wallet's wording should win.

GigaHierz (approved):
Re-verified: all three Sepolia addresses independently confirmed on-chain (symbols and decimals match, including the USD₮ U+20AE spelling on mainnet), the minipay.mdx:115 snippet now matches the exported TOKENS_BY_CHAIN, and the USDm relabel is correctly left as a maintainer decision — approving with the label as-is; the token reference table update rides in #444 as scoped. One cosmetic nit, non-blocking: the unknown-chain message renders awkwardly if chainId is ever undefined while connected.

Issue #387, opened 2026-06-08
ai-chat template: fails to scaffold (template path bug) + undeployable out of the box


GigaHierz:
## Summary

The `ai-chat` template is broken in two ways: (1) it **fails to scaffold at all** due to a template-path bug, and (2) once you work around that, it is **not deployable out of the box** (Vercel blocks it on a vulnerable Next.js, the default install fails on peer deps, and the build hard-fails without a database). Other templates (e.g. `minipay`) scaffold fine — only `ai-chat` hits #1.

Reproduced with `@celo/celo-composer@2.4.13`, Node 20.19, macOS.

---

## Issue 1 — `ai-chat` fails to scaffold (ENOENT)

```bash
npx @celo/celo-composer@latest create myapp -t ai-chat -y
```
```
- Generating project structure...
Error generating project: [Error: ENOENT: no such file or directory, lstat
'.../@celo/celo-composer/dist/templates/ai/chat-template']
✖ Failed to generate project
```

**Root cause.** Template files ship at the **package root** (`<pkg>/templates/ai/chat-template`), but the resolver looks under `<pkg>/dist/templates/...`. In `src/utils/paths.ts` → `getTemplatesPath()`, the `isInDist` branch returns:

```js
templatesPath = join(currentDir, "..", "templates");
```

That assumes the caller sits directly in `dist/`. It's correct for `dist/plopfile.js` (`dist/../templates` → `<root>/templates` ✓ — which is why the plop-based templates work), but wrong for callers one level deeper. The `ai`/`chat-template` copy runs through `dist/generators/plop-generator.js`:

```js
const templatesPath = getTemplatesPath(import.meta.url);          // dist/generators → dist/templates ✗
const sourcePath = path.join(templatesPath, "ai", "chat-template");
await safeCopyTemplate(sourcePath, projectPath);
```

so it resolves to the non-existent `dist/templates/ai/chat-template`.

**Fix.** Make `getTemplatesPath()` depth-independent (resolve to package root regardless of caller location):

```ts
const isInDist = /[\\/]dist([\\/]|$)/.test(currentDir);
if (isInDist) {
  const root = currentDir.split(/[\\/]dist([\\/]|$)/)[0];
  return join(root, "templates");
}
```

Verified locally: symlinking `dist/templates → ../templates` makes `-t ai-chat` scaffold successfully, confirming the path is the only blocker. (Also filed from the docs side as celo-org/celopedia-skills#23.)

---

[...]

PR #437, opened 2026-08-12
fix(templates): stop the ai-chat app importing test mocks in production

### LINKED ISSUE #424: ai-chat template hygiene: production providers.ts imports ./models.test, 1 MB tsconfig.tsbuildinfo committed, React 19 RC with @types/react 18
## Summary

Build/code hygiene problems in the vendored `templates/ai/chat-template` app (distinct from the scaffolding/deploy problems in #387):

1. **Test mocks imported into production code.** `lib/ai/providers.ts:16` imports from `./models.test` at module top level. Every scaffolded app bundles test-mock code into its production build, and any tooling that strips test files breaks the import.

2. **A 1 MB build artifact is committed.** `templates/ai/chat-template/tsconfig.tsbuildinfo` (~1,037 KB) is checked in and references absolute `node_modules` paths from a local machine. The template's own `.gitignore` doesn't cover `*.tsbuildinfo` (the base template's does). It is stripped from the npm tarball by the prepack script, but it bloats the repo and every git clone.

3. **React version/type mismatch.** `package.json` pins `react`/`react-dom` to the RC snapshot `19.0.0-rc-45804af1-20241021` while declaring `@types/react: ^18` / `@types/react-dom: ^18` — types and runtime are a major version apart. `eslint-config-next` is pinned to `14.2.5` against `next 15.3.6`.

4. **Stale default model IDs.** `lib/ai/models.ts` defaults to `gemini-1.5-flash-latest` / `gemini-1.5-pro-latest` / `gpt-4o-mini` — worth refreshing along with the AI SDK beta pins (`ai 5.0.0-beta.6`, `@ai-sdk/react 2.0.0-beta.6`).

## Suggested fix

Move the mock models into a non-test module imported by both `providers.ts` and the tests; delete `tsconfig.tsbuildinfo` and gitignore it; align `@types/react*` with the shipped React (or move to stable 19 + Next 15 types); bump default model IDs when touching the file.

ghozzza:
Closes #424, items 1–3. Item 4 is deliberately not included — reasoning at the bottom.

## 1. Test mocks in the production bundle

`lib/ai/providers.ts:16` imported four `MockLanguageModelV2` instances from `./models.test` at module top level, so every scaffolded app pulled test-only code into its production build, and any tooling that strips test files broke the import outright.

Renamed to `models.mock.ts`. Worth noting it was never actually a test file — `playwright.config.ts` only globs `tests/e2e/*.test.ts` and `tests/routes/*.test.ts`, so nothing ever ran it. It only *looked* like one to everything that reads filenames, which is exactly the hazard. `providers.ts` is its only importer, so the rename is the whole change.

## 2. The committed build artefact

`tsconfig.tsbuildinfo` — 1,037,467 bytes of absolute `node_modules` paths from someone's machine — was tracked, and the template's own `.gitignore` did not cover it. Deleted and gitignored.

It never reached users: `scripts/prepack-prune.cjs` strips it from the tarball and `src/utils/safe-copy.ts` excludes it when copying into a generated project. Both guards stay — they are the right defence if someone regenerates it locally. The cost was to the repo and to every clone, which is reason enough.

## 3. React RC with React 18 types

```
react / react-dom : 19.0.0-rc-45804af1-20241021     ← an October 2024 RC snapshot
@types/react      : ^18                             ← a major version behind the runtime
eslint-config-next: 14.2.5                          ← against next 15.3.6
```

Moved to stable `^19` with matching `^19` types, and `eslint-config-next` to `15.3.6` so it matches the shipped Next. A template pinning a dated RC is worse than a template pinning stable, and the type/runtime split is a major version wide.

## 4. Model IDs — not changed, on purpose

`gemini-1.5-flash-latest`, `gemini-1.5-pro-latest` and `gpt-4o-mini` are still there.

Bumping them means asserting that specific third-party model identifiers exist and behave, which I cannot verify from here, and I would rather leave a working older default than write an ID I have not confirmed. It is also a cost-and-capability call rather than hygiene — which model an ai-chat scaffold should default to is a product decision. Happy to take a direction if you have one.

## Verification, and its caveat

**The ai-chat template cannot be scaffolded on `main` at all** — `ENOENT: .../dist/templates/ai/chat-template`, which is #387 and fixed in #436. So this was verified with that fix applied locally on top:

```
lib/ai/            entitlements.ts  models.mock.ts  models.ts  prompts.ts  providers.ts  tools
providers.ts       from "./models.mock"
models.test        no references anywhere in the generated project
tsconfig.tsbuildinfo  absent
react              ^19.0.0      @types/react  ^19
eslint-config-next 15.3.6       next          15.3.6
```

**#436 should land first**, otherwise there is no way to confirm any of this end to end.


GigaHierz (approved):
Verified earlier: providers.ts is the only importer of the renamed mock module, the isTestEnvironment branch keeps all four model exports, and the react/next version bumps check out against registry peer ranges. Note for a follow-up (non-blocking): models.mock.ts imports @/tests/prompts/utils, so production code still depends on the tests/ directory.

PR #438, opened 2026-08-12
fix(build): stash pruned files outside the package, not inside it

### LINKED ISSUE #413: prepack stashes pruned files inside the packed directory — .next and tsbuildinfo re-enter the npm tarball under .prune-stash/
## What happens

The `prepack` prune mechanism stashes the files it wants to exclude *inside the directory being packed*, so they can re-enter the published tarball under `.prune-stash/`.

Verified locally: with a `.next/` directory present in `templates/ai/chat-template/` (as any developer who ran the template locally would have), `pnpm pack` produces:

```
package/templates/ai/chat-template/.prune-stash/manifest.json
package/templates/ai/chat-template/.prune-stash/tsconfig.tsbuildinfo
package/templates/ai/chat-template/.prune-stash/.next/marker.txt
```

The prune achieves the opposite of its intent: files moved "out of the way" are shipped anyway, just under a different path.

## Root cause

`scripts/prepack-prune.cjs:7-34` moves `.git`, `.next`, `node_modules`, `tsconfig.tsbuildinfo`, `.env`, `.env.local` into `templates/ai/chat-template/.prune-stash/` — but `templates` is in the package `files` list, and npm/pnpm's implicit ignore rules only cover some of these names at arbitrary depth (`.git`, `node_modules`; in our test pnpm also dropped the stashed `.env`, but `.next` and `tsconfig.tsbuildinfo` shipped). Whether an env file leaks may vary by package-manager version — the mechanism itself is unsafe.

Secondary issues:

- If `postpack` never runs (aborted pack, CI kill), `.prune-stash/` persists in the working tree and is not in `.gitignore` (which lists only `node_modules`, `.env`, `dist`).
- The prune list has drifted from `src/utils/safe-copy.ts:8-9` (which also excludes `docs`), and neither list covers `tests/`, `playwright.config.ts`, `.github/`, `.vscode/` — all currently shipped in the tarball.

## Fix

Stash outside the package root (e.g. `os.tmpdir()`), or generate the tarball from a clean export. Add `.prune-stash` to `.gitignore` regardless.

Found in @celo/celo-composer 2.4.13 (main @ 710dd89), tested with pnpm 10.13.

ghozzza:
Closes #413.

## The mechanism was self-defeating

`prepack-prune.cjs` moved the files it wanted to exclude into `templates/ai/chat-template/.prune-stash/` — **inside the tree being packed**. `templates` is in the package `files` list, so npm's implicit ignores only caught some of them at that depth, and the rest shipped under the stash path instead of their original one.

**Measured on `main`**, with a `.next/` present in the template as any developer who ran it locally would have:

```
$ pnpm pack && tar -tzf celo-celo-composer-2.4.13.tgz | grep -c prune-stash
3869
```

Including `package/templates/ai/chat-template/.prune-stash/tsconfig.tsbuildinfo`.

pnpm 10.13 happened to drop the stashed `.env`. That is luck, not design — the issue already notes it varies by package manager, and nothing in the mechanism prevents an env file from shipping.

## The fix

Stash in `os.tmpdir()`, keyed on a hash of the package root:

```js
const STASH_KEY = crypto.createHash('sha256').update(ROOT).digest('hex').slice(0, 12);
const STASH_DIR = path.join(os.tmpdir(), `celo-composer-prune-${STASH_KEY}`);
```

Derived identically in both scripts, so they agree without passing state, and keyed on ROOT so two checkouts packing concurrently cannot collide.

`postpack` also drains and removes a legacy `.prune-stash/` if an older version of these scripts left one in the tree — otherwise a stale one sits there until someone notices. `.gitignore` covers it regardless, per the issue.

## One thing beyond the issue: both scripts now fail rather than warn

```js
} catch (err) {
  console.error('[prepack] prune step FAILED — refusing to pack:', err);
  process.exitCode = 1;
}
```

The previous `console.warn('...(continuing)')` is the wrong default for a step whose entire job is keeping `.env` and `.env.local` out of a published tarball. A warning scrolls past and the publish proceeds. Same treatment on `postpack`, where a failure means the working tree is missing whatever was stashed — it now says so and prints the stash path so it can be recovered by hand.

## Verified

Same fixtures — `.next/`, `.env`, `tsconfig.tsbuildinfo` all present before packing:

| | main | this branch |
|---|---|---|
| `.prune-stash` entries in tarball | 3,869 | **0** |
| `chat-template/.next` entries | present | **0** |
| `tsbuildinfo` | present | **0** |
| stray `.env` | 0 (pnpm luck) | **0** |

And all three are restored to the working tree afterwards, with the temp stash removed.

## Not included

The issue also notes the prune list has drifted from `src/utils/safe-copy.ts`, and that `tests/`, `playwright.config.ts`, `.github/` and `.vscode/` currently ship in the tarball. Deciding what *should* ship is a package-size call rather than a correctness one, so I have left it — happy to take it as a follow-up if you want the two lists reconciled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 node_modules
 .env
 dist
+# leftover from an interrupted pnpm pack (older versions stashed here)
+.prune-stash/
--- a/scripts/postpack-restore.cjs
+++ b/scripts/postpack-restore.cjs
@@ -2,26 +2,53 @@
 /* eslint-disable no-console */
 const fs = require('fs-extra');
 const path = require('path');
+const os = require('os');
+const crypto = require('crypto');
 
 const ROOT = path.join(__dirname, '..');
 const TEMPLATE_PATH = path.join(ROOT, 'templates', 'ai', 'chat-template');
-const STASH_DIR = path.join(TEMPLATE_PATH, '.prune-stash');
+
+// Must match prepack-prune.cjs exactly — same derivation, no shared state.
+const STASH_KEY = crypto.createHash('sha256').update(ROOT).digest('hex').slice(0, 12);
+const STASH_DIR = path.join(os.tmpdir(), `celo-composer-prune-${STASH_KEY}`);
+
+// Where the stash used to live, inside the packed tree. Cleaned up if an older
+// version of these scripts left one behind, so a stale directory cannot be
+// picked up by a later pack.
+const LEGACY_STASH_DIR = path.join(TEMPLATE_PATH, '.prune-stash');
 
 (async () => {
   try {
+    if (await fs.pathExists(LEGACY_STASH_DIR)) {
+      const legacyManifest = path.join(LEGACY_STASH_DIR, 'manifest.json');
+      if (await fs.pathExists(legacyManifest)) {
+        for (const item of await fs.readJson(legacyManifest)) {
+          const destPath = path.join(TEMPLATE_PATH, item.rel);
+          if (await fs.pathExists(item.dest)) {
+            await fs.ensureDir(path.dirname(destPath));
+            await fs.move(item.dest, destPath, { overwrite: true });
+            console.log(`[postpack] restored ${item.rel} from the legacy stash`);
+          }
+        }
+      }
+      await fs.remove(LEGACY_STASH_DIR);
+    }
+
     if (!(await fs.pathExists(STASH_DIR))) return;
     const manifestPath = path.join(STASH_DIR, 'manifest.json');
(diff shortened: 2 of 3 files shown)


GigaHierz (approved):
Verified empirically with fixtures: tarball contains zero .prune-stash/.next/tsbuildinfo/.env entries, restore is symmetric, STASH_KEY stable through /tmp symlinks. One pre-existing edge worth a follow-up issue: crash-before-postpack followed by a second pack overwrites the manifest with [] and the stash cleanup deletes the stranded files — prepack should refuse to run (or merge) when a manifest already exists.

PR #440, opened 2026-08-12
fix(cli): let a flag bypass its own prompt, not all of them

### LINKED ISSUE #411: Any single create flag silently disables ALL interactive prompts, contradicting the documented per-option bypass
## What happens

Passing *any single* flag to `create` — even just a description — silently disables **all** interactive prompts and applies defaults for everything else:

```bash
npx @celo/celo-composer@2.4.13 create myapp -d "hello"
# → no prompts at all; generates basic + rainbowkit + hardhat
```

Verified: the command above produced `apps/web` + `apps/contracts` (hardhat) with zero questions asked.

## Why this is a problem

- `docs/cli-reference/create-command.mdx:21` documents the opposite: "Using any of these flags will bypass the interactive prompts **for those options**" — i.e. per-option bypass, remaining options still prompted.
- A user adding `-d` to the interactive flow gets rainbowkit + hardhat without ever being asked.
- It makes `-y/--yes` ("skip all prompts") effectively redundant, since every other flag already does that.

## Root cause

`src/commands/create.ts:49-57` computes `hasCliOptions` as true if *any* of `description | template | templateType | walletProvider | contracts | skipInstall | yes` is set, and line 59-60 switches the entire prompt block off when it is.

## Suggested fix

Prompt for exactly the options not provided (inquirer `when:` per question keyed on the specific flag), and reserve `-y` for "accept defaults for everything unanswered". That matches the documented behavior.

ghozzza:
Closes #411.

## The bug

```bash
npx @celo/celo-composer create myapp -d "hello"
# no prompts at all → basic + rainbowkit + hardhat
```

`hasCliOptions` was true if *any* of seven flags was set, and that switched off the whole prompt block. So adding a description silently chose a wallet provider and a contract framework, and `-y` was redundant — every other flag already did what it does.

`docs/cli-reference/create-command.mdx:21` documents the opposite: a flag bypasses the prompts **"for those options"**.

**The per-question `when:` clauses already implement that**, keyed on the specific flag. They were never reached. So the fix is that only `-y` skips.

## The second bug, which the first was hiding

`-t` populates `options.template`. Only the long `--template-type` would set `options.templateType`, and **four reads checked only the latter**:

| line | read | consequence once prompts run |
|---|---|---|
| template prompt `when:` | `!options.templateType` | re-asks for a template already chosen |
| wallet `when:` | `options.templateType \|\| answers.templateType` | asks minipay for a wallet provider it forces itself |
| contracts `when:` | same | same shape |
| **`finalTemplateType`** | `options.templateType \|\| answers.templateType \|\| "basic"` | **`-t minipay` generates a `basic` project** |

That last one is the serious one, and it is only reachable *after* fixing the bypass — the auto-mode branch computed it correctly, so nothing was wrong while every flag went down that path. Removing the bypass without this would have broken `-t` outright.

Normalised into a single `cliTemplateType` at the top rather than patching four sites, so a fifth read cannot repeat it.

## Verified

Prompts driven through stdin, recording which questions each invocation asks:

| invocation | asks |
|---|---|
| `-d hello` | template, wallet — **not** description |
| `-t minipay` | description, contracts — **not** template, **not** wallet |
| `-t basic` | description, wallet — **not** template |
| `-y` | nothing |

And that `-t` still reaches generation, checked on a file unique to each template:

```
-t minipay            → apps/web/src/components/user-balance.tsx   PRESENT
-t farcaster-miniapp  → apps/web/src/contexts/miniapp-context.tsx  PRESENT
-t basic              → apps/contracts/hardhat.config.ts           PRESENT
```

`tsc --noEmit` clean. `-y` scaffolds exactly as before.

## Conflict note

~~This touches `src/commands/create.ts`, which #393 and #430 also modify — different hunks … so they should merge in any order.~~ **Stale, and wrong now.** Re-checked against the current heads: **#393 conflicts.**

```
$ git merge --no-ff x393
CONFLICT (content): Merge conflict in src/commands/create.ts
```

The heads moved after this note was written — #393 now rewrites the same auto-mode block and the same `when:` clauses this PR normalises. **#430 still merges clean.**

**Recommended order, which is the reviewer's: this PR first, #393 rebases last.** On that rebase #393 needs two things, both known: `'x402'` added to #442's `.choices()` list, and its x402 `when:` conditions re-threaded onto `cliTemplateType` — the current clauses read `options.templateType`, which is exactly the field this PR normalises away because it misses `-t`.

## Behavioural break — worth a release note

**This is a deliberate break for anything that relied on the old any-flag auto-mode.** `create app -t basic --skip-install` in CI used to run silently to completion; it now stops at the description prompt. That is the documented behaviour restored (#411), not a regression, but it changes what an existing script does with no error message to explain it.

The release-note line belongs in the CHANGELOG, which **#446** currently rewrites end to end — adding it here would collide with that PR for no benefit, so it goes in #446's `[Unreleased]` section and is noted on both.

### DIFF
--- a/docs/cli-reference/create-command.mdx
+++ b/docs/cli-reference/create-command.mdx
@@ -43,8 +43,10 @@ npx @celo/celo-composer@latest create my-awesome-dapp
 
 Create a new project with specific configurations by using flags. This is useful for scripting and automation.
 
+**A flag only skips its own question.** `--template` answers the template prompt and nothing else, so anything you leave out is still asked. In a script or a CI job there is nobody to answer, and the command stops there — so **add `-y` whenever the run has no terminal attached**. It accepts the defaults for every question your flags did not already answer, and your flags still win where you gave them.
+
 ```bash
-npx @celo/celo-composer@latest create --template minipay
+npx @celo/celo-composer@latest create my-minipay-app --template minipay -y
 ```
 
 This command creates a new project using the Minipay template, which is optimized for integration with the Minipay mobile wallet. It comes with a pre-configured wallet connection and a mobile-first UI.
@@ -53,14 +55,16 @@ This command creates a new project using the Minipay template, which is optimize
 npx @celo/celo-composer@latest create my-farcaster-app \
   --template farcaster-miniapp \
   --wallet-provider thirdweb \
-  --contracts none
+  --contracts none \
+  -y
 ```
 
 ```bash
 npx @celo/celo-composer@latest create my-ai-chat \
   --template ai-chat \
   --contracts none \
-  --wallet-provider none
+  --wallet-provider none \
+  -y
 ```
 
 The `ai-chat` template is a standalone Next.js app copied as-is (no Handlebars rendering). Wallet and contracts prompts are skipped by default.
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -44,23 +44,28 @@ export async function createCommand(
   try {
     console.log(chalk.blue.bold("\n🟨 Welcome to Celo Composer CLI!\n"));
 
-    // If --yes flag is provided, skip all prompts and use defaults/flags
-    // Check if any CLI options are provided (auto-mode)
-    const hasCliOptions = !!(
-      options.description ||
-      options.template ||
-      options.templateType ||
-      options.walletProvider ||
-      options.contracts ||
-      options.skipInstall ||
-      options.yes
-    );
+    // `-t` populates options.template; only the long `--template-type` sets
+    // templateType. Several reads below checked just one of the two, which was
+    // harmless while any flag skipped the prompts entirely and is not once they
+    // run: the template question would re-ask, and finalTemplateType would fall
+    // through to "basic" for a `-t minipay` that had no -y. Normalised once.
+    const cliTemplateType = options.template || options.templateType;
 
+    // Only -y skips the prompts.
+    //
+    // This used to skip them whenever ANY flag was present, so
+    // `create myapp -d "hello"` silently produced basic + rainbowkit + hardhat
+    // without asking a single question — and made -y redundant, since every
+    // other flag already did what it does. The docs describe the opposite:
+    // a flag bypasses the prompt "for those options", leaving the rest asked.
+    //
+    // The per-question `when:` clauses below already implement that, keyed on
+    // the specific flag. They were simply never reached.
     const answers: PromptAnswers =
-      options.yes || hasCliOptions
+      options.yes
         ? ((): PromptAnswers => {
             const autoTemplateType =
-              options.template || options.templateType || "basic";
+              cliTemplateType || "basic";
             const autoWallet = options.walletProvider
               ? options.walletProvider
               : autoTemplateType === "farcaster-miniapp" ||
@@ -112,7 +117,10 @@ export async function createCommand(
                 { name: "AI Agent Chat App", value: "ai-chat" },
               ],
               default: "basic",
-              when: !options.templateType,
+              // `-t` populates options.template; only the long form sets
+              // templateType. Checking one of the two asks again for a template
+              // the user already chose.
+              when: !cliTemplateType,
             },
             {
               type: "list",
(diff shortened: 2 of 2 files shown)


ghozzza:
Cross-PR note, found by merging all 24 open PRs into one tree and running everything rather than testing this branch alone.

**This PR breaks #393's test suite until a one-line change lands there.** #393's fixture invokes the CLI as:

```
tsx src/index.ts create x402-fixture -t x402 --skip-install
```

no `-y`. That works today because any flag disables every prompt — which is the bug this PR fixes. Afterwards the fixture blocks on an interactive question with no TTY:

```
Error: readline was closed
7 tests failed
```

**Already fixed on #393's branch** (`-y` added to the fixture, with the reason in a comment), so the two merge in either order now. Flagging it because it is the kind of breakage CI on this branch cannot show — the test that breaks lives in another PR.

GigaHierz (changes requested):
The code change is correct and the `cliTemplateType` normalization is genuinely necessary (without it, `-t minipay` with a skipped prompt would fall through to `"basic"`). Two items before merge:

1. **This PR changes the semantics that the docs' own examples rely on, without updating them.** `docs/cli-reference/create-command.mdx` (~lines 42–48, "Non-Interactive Mode (with flags) … useful for scripting and automation") shows `npx @celo/celo-composer@latest create --template minipay` with no `-y`. Under this PR that exact command now prompts for description and contracts — in the CI/scripting context the section recommends, it hangs. The examples need `-y` added (or the section reworded), and since this PR is what changes the behavior, the doc fix belongs here.

2. **The conflict note is stale.** The body says #393 touches "different hunks — they should merge in any order," but a `git merge-tree` check of the current heads shows a real conflict in `src/commands/create.ts` (both rewrite the auto-mode wallet/contract block and the `when:` clauses). Suggest updating the note; recommended order is this PR first, #393 rebasing last.

One more callout worth adding to the body: this is a deliberate behavioral break for any script that relied on the old any-flag auto-mode (e.g. `create app -t basic --skip-install` in CI now prompts). It's the documented-behavior fix from #411, but it deserves a release-note line so downstream users aren't surprised.

ghozzza:
All three addressed at `61231f6`. One of them was bigger than the review said.

## The docs examples — it was three of four, not one

You named `create --template minipay`. Running the whole section headless found two more, failing on a different prompt:

| example | headless result before |
|---|---|
| `create --template minipay` | blocks on **project name** |
| `create my-farcaster-app --template farcaster-miniapp --wallet-provider thirdweb --contracts none` | blocks on **description** |
| `create my-ai-chat --template ai-chat --contracts none --wallet-provider none` | blocks on **description** |
| `create my-default-project --yes` | fine |

All three now carry `-y`, and the section states the rule once rather than four times: **a flag skips only its own question, so anything left unspecified is still asked** — which is precisely what this PR changes, so the sentence belongs here.

Checked that `-y` does not flatten the explicit flags before writing that down: `-t farcaster-miniapp --wallet-provider thirdweb --contracts none -y` still yields the farcaster deps, thirdweb, and no `apps/contracts`.

**Swept the rest of `docs/` and `README.md` rather than fixing only the file you named.** Every other `create` example is either interactive by design — `quickstart.mdx` says *"The CLI will guide you through a series of prompts"* and already carries a `--yes` pro-tip, `installation.mdx` is the bare `create` — or lives in a file another open PR is rewriting (`README.md` in #393/#443/#444, `minipay.mdx` in #444, `quickstart.mdx` and `foundry.mdx` in #443). Editing those here would conflict for no gain; nothing there is broken in the scripting sense.

One caveat on the ai-chat line: it still fails headless, but on `ENOENT … dist/templates/ai/chat-template` — that is #387/#436, not a prompt. The `-y` is right; the example is unverifiable until #436 lands.

## Conflict note — corrected, and you were right

Re-ran it rather than taking the note's word:

```
$ git merge --no-ff x393
CONFLICT (content): Merge conflict in src/commands/create.ts
```

The note predates both heads moving. Body now says so, keeps **#430 merges clean**, and records the order as yours: **this first, #393 rebases last**, re-threading its `when:` clauses onto `cliTemplateType` since the current ones read `options.templateType` — the exact field this PR removes for missing `-t`.

## Release note

Added the break to the body as a callout. The CHANGELOG line itself goes in **#446**, which rewrites that file end to end — putting it here would collide with it for no benefit. Noted on both PRs so it cannot fall between them.

GigaHierz (approved):
Re-verified at 61231f6: all three docs examples now carry -y (plus the name arg), the conflict note matches an independent merge-tree run, and the behavioral-break callout is in place. Accepting the CHANGELOG line living in #446 — but note that creates a merge-order dependency: this PR must land before #446, whose [Unreleased] entry describes this behavior.

PR #442, opened 2026-08-12
fix(cli): reject unknown flag values, and emit an npm-legal package name

### LINKED ISSUE #412: create accepts invalid input: unknown -t values silently generate a bare project; uppercase names yield invalid npm package names
## What happens

`create` accepts invalid input without complaint, in two ways:

1. **Unknown template values are silently accepted.**
   ```bash
   npx @celo/celo-composer@2.4.13 create app-bogus -t bogus --skip-install -y
   # → succeeds; generates a bare base project with no wallet/miniapp overlays
   ```
   `-t/--template` (`src/index.ts:19`) and the other choice flags (`--wallet-provider`, `-c`) are never validated against their allowed values — the string flows straight into Plop, where every `{{#if (eq templateType ...)}}` condition is simply false. The user gets a project that doesn't match anything documented, with no warning.

2. **Uppercase project names produce invalid npm package names.**
   ```bash
   npx @celo/celo-composer@2.4.13 create MyApp --skip-install -y
   grep '"name"' MyApp/package.json   # → "name": "MyApp"
   ```
   `src/utils/validation.ts:9` allows `[a-zA-Z0-9_-]`, and the value is interpolated verbatim into `templates/base/package.json.hbs:2`. npm rejects new package names containing uppercase letters, so tooling that validates the manifest (`npm pkg fix`, publishing, some CI) chokes on the generated project.

## Suggested fix

- Validate `--template`, `--wallet-provider`, and `--contracts` against their known values and exit with the list of valid choices on mismatch.
- Either reject uppercase project names or lowercase the value used for the package `name` field (keeping the directory name as typed). A `kebabCase` handlebars helper is already registered in `src/plopfile.ts` and could be applied at the usage site.

ghozzza:
Closes #412.

## 1. Unknown flag values were accepted

```
$ create app-bogus -t bogus --skip-install -y
→ succeeded, generated a bare base project
```

The string went straight into Plop, where every `{{#if (eq templateType "…")}}` is simply false, so the user got a project matching nothing in the docs and no indication anything was wrong.

Declared with commander's `.choices()` instead, which rejects before anything is written:

```
error: option '-t, --template <type>' argument 'bogus' is invalid.
       Allowed choices are basic, farcaster-miniapp, minipay, ai-chat.
error: option '--wallet-provider <provider>' argument 'nope' is invalid.
       Allowed choices are rainbowkit, thirdweb, none.
error: option '-c, --contracts <framework>' argument 'nope' is invalid.
       Allowed choices are hardhat, foundry, none.
```

No directory is created in any of those cases, and `-t minipay -c foundry` still scaffolds normally.

## 2. Uppercase names produced invalid npm manifests

`create MyApp` wrote `"name": "MyApp"` into `package.json`. npm refuses uppercase in new package names, so `npm pkg fix`, publishing, and some CI steps fail on a project the tool just generated.

Took the second option the issue offers — **lowercase the manifest field, keep the directory as typed** — because the directory name is the user's choice and not npm's business, and rejecting it outright would break anyone who wants `MyApp` as a folder. The `kebabCase` helper was already registered in `plopfile.ts`.

| `create …` | directory | manifest `name` |
|---|---|---|
| `MyApp` | `MyApp` | `my-app` |
| `my-app` | `my-app` | `my-app` |
| `celo-demo` | `celo-demo` | `celo-demo` |
| `app2` | `app2` | `app2` |

~~Only `templates/base/package.json.hbs` interpolates the project name into a `name` field, so that is the whole change.~~ **Wrong — corrected at `e1131c9`.** That was true of the `.hbs` files and false of the tool. `src/generators/plop-generator.ts` copies the ai-chat template verbatim and patches its `package.json` in JS, never reaching Plop, so the helper never ran there.

**The sweep, done properly this time, and reproducible:**

```bash
grep -rn '"name":.*projectName' templates/     # 1 hit  templates/base/package.json.hbs:2
grep -rn '\.name\s*=' src/ --include='*.ts'    # 1 hit  plop-generator.ts:64
find templates -name package.json               # 1 hit  ai/chat-template  <- the raw copy
```

Two write sites, not one, and the third command is the one that would have caught it: a template shipping a real `package.json` rather than a `.hbs` cannot be found by grepping for the interpolation.

## ⚠️ Merge-order dependency with #393

**#393 adds `x402` as a template type.** The choices list here does not include it, because on `main` that template does not exist — accepting `-t x402` today would produce exactly the silent bare project this PR removes.

So whichever lands second needs one line:

- **#412 first, then #393** → add `'x402'` to the `-t` choices in #393.
- **#393 first, then #412** → include `'x402'` in the list here before merging.

Flagging it because CI will not catch it: an omitted choice is a rejected flag, not a failing build. I will make the edit on whichever side is second as soon as the other lands.

### DIFF
--- a/src/generators/plop-generator.ts
+++ b/src/generators/plop-generator.ts
@@ -3,6 +3,7 @@ import nodePlop from "node-plop";
 import path from "path";
 import { safeCopyTemplate } from "../utils/safe-copy.js";
 import { getTemplatesPath, getPlopfilePath } from "../utils/paths.js";
+import { toPackageName } from "../utils/validation.js";
 
 // TypeScript compilation handles TS support - no runtime tsx needed
 
@@ -62,7 +63,9 @@ export class TemplateGenerator {
         const pkgPath = path.join(projectPath, "package.json");
         if (await fs.pathExists(pkgPath)) {
           const pkg = JSON.parse(await fs.readFile(pkgPath, "utf8")) as Record<string, unknown>;
-          (pkg as { name?: string }).name = projectName;
+          // This path never reaches Plop, so the kebabCase helper never runs —
+          // normalise here through the same function the helper uses.
+          (pkg as { name?: string }).name = toPackageName(projectName);
           // Normalize version for new project scaffolds
           (pkg as { version?: string }).version = "0.1.0";
           await fs.writeFile(pkgPath, JSON.stringify(pkg, null, 2) + "\n");
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { Command } from 'commander';
+import { Command, Option } from 'commander';
 import { createRequire } from 'node:module';
 import { createCommand } from './commands/create.js';
 import chalk from 'chalk';
@@ -24,9 +24,32 @@ program
   .description('Create a new Celo project')
   .argument('[project-name]', 'Name of the project')
   .option('-d, --description <description>', 'Project description')
-  .option('-t, --template <type>', 'Template type (basic, farcaster-miniapp, minipay, ai-chat)')
-  .option('--wallet-provider <provider>', 'Wallet provider (rainbowkit, thirdweb, none)')
-  .option('-c, --contracts <framework>', 'Smart contract framework (hardhat, foundry, none)')
+  // .choices() rather than a bare string, so an unknown value is rejected with
+  // the allowed list instead of flowing into Plop — where every
+  // `{{#if (eq templateType ...)}}` is simply false and the user silently gets a
+  // bare base project that matches nothing documented.
+  .addOption(
+    new Option('-t, --template <type>', 'Template type').choices([
+      'basic',
+      'farcaster-miniapp',
+      'minipay',
+      'ai-chat',
+    ])
+  )
+  .addOption(
+    new Option('--wallet-provider <provider>', 'Wallet provider').choices([
+      'rainbowkit',
+      'thirdweb',
+      'none',
+    ])
+  )
+  .addOption(
+    new Option('-c, --contracts <framework>', 'Smart contract framework').choices([
+      'hardhat',
+      'foundry',
+      'none',
+    ])
+  )
   .option('--skip-install', 'Skip package installation')
   .option('-y, --yes', 'Skip all prompts and use defaults')
   .action(createCommand);
--- a/src/plopfile.ts
+++ b/src/plopfile.ts
@@ -1,6 +1,7 @@
 import { NodePlopAPI } from "plop";
 import path from "path";
 import { getTemplatesPath } from "./utils/paths.js";
+import { toPackageName } from "./utils/validation.js";
 
 interface PlopData {
   projectName: string;
@@ -261,12 +262,7 @@ export default function (plop: NodePlopAPI): void {
   });
 
   // Add helper functions for templates
-  plop.setHelper("kebabCase", (text: string) => {
-    return text
-      .replace(/([a-z])([A-Z])/g, "$1-$2")
-      .replace(/\s+/g, "-")
-      .toLowerCase();
-  });
+  plop.setHelper("kebabCase", (text: string) => toPackageName(text));
 
   plop.setHelper("camelCase", (text: string) => {
     return text
--- a/src/utils/validation.ts
+++ b/src/utils/validation.ts
@@ -1,3 +1,25 @@
+/**
+ * The project name as it belongs in a package.json `name` field.
+ *
+ * npm has refused uppercase in new package names since 2017, so the directory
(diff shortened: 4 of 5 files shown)


ghozzza:
Confirmed by simulation, and the resolution is now known.

Merging all 24 open PRs into one tree produces exactly the conflict this PR's body predicted — `src/index.ts`, between the `.choices()` list here and #393's `-t` description. The resolution is one line:

```ts
new Option('-t, --template <type>', 'Template type').choices([
  'basic',
  'farcaster-miniapp',
  'minipay',
  'ai-chat',
  'x402',        // ← whichever of #442 / #393 lands second adds this
])
```

Also worth knowing: with all 24 merged and that line added, `-t x402` scaffolds and `-t bogus` is still rejected with the allowed list, so the two changes compose correctly.

GigaHierz (changes requested):
The choices-validation and the `{{kebabCase projectName}}` change are both correct, and the merge-order plan with #393 is fine. One gap blocks the headline claim:

**The ai-chat raw-copy path still emits the npm-invalid name.** `src/generators/plop-generator.ts:61` sets `pkg.name = projectName` verbatim on the copied package.json — this path bypasses Plop entirely, so the `kebabCase` helper never runs. `create MyApp -t ai-chat` still writes `"name": "MyApp"`.

The PR body's statement that "only `templates/base/package.json.hbs` interpolates the project name into a `name` field, so that is the whole change" is therefore not accurate — that's true of the `.hbs` files, but this JS site is a fourth interpolation point. It's a one-line fix (apply the same lowercasing/kebab logic in `plop-generator.ts`), and the body claim should be corrected alongside it.

Reminder from your own coordination note: whichever of this PR and #393 lands second must add `'x402'` to the `.choices()` list.

ghozzza:
Fixed at `e1131c9`, and the body claim is struck through with the sweep that should have been run.

**The fix.** Both paths now call one `toPackageName` in `utils/validation.ts` rather than each rendering the name themselves — the helper delegates to it, and the raw-copy path calls it directly. Put there deliberately: `validateProjectName` sits in the same file and has already constrained the input to `[a-zA-Z0-9_-]` starting with an alphanumeric by the time either runs, which is *why* lowercasing is sufficient to produce a legal name. That coupling is now written down next to both, so loosening the charset cannot silently invalidate the claim.

**Verified by generating, not by reading:**

| `create …` | directory | manifest `name` |
|---|---|---|
| `MyApp -t ai-chat` | `MyApp` | `my-app` |
| `MyApp2 -t basic` | `MyApp2` | `my-app2` |
| `My_App3 -t minipay` | `My_App3` | `my_app3` |

Directory keeps the user's casing; only the manifest field is normalised, which is what the issue asked for.

**And it bites.** Reverting just that line and regenerating puts `"name": "MyApp"` straight back, so the assertion is load-bearing rather than incidentally true.

**One thing worth flagging: this could not be verified on this branch alone.** `-t ai-chat` fails to scaffold at all on current `main` — `ENOENT … dist/templates/ai/chat-template`, which is #387/#436. So the run above is against this branch **with #436 merged in**. Not a change to the merge order, but it does mean the ai-chat half of this PR is only observable once #436 lands.

Noted on the coordination point too: whichever of this and #393 lands second still adds `'x402'` to the `.choices()` list.

GigaHierz (approved):
Re-verified at e1131c9: the ai-chat raw-copy path now goes through the shared toPackageName (executed against MyApp/My_App3/MyAPP — matches the kebabCase helper byte-for-byte, and the helper now delegates to the same function so the paths cannot drift), directory casing preserved, body corrected. Reminder: whichever of this and #393 merges second adds 'x402' to the choices list.

GigaHierz:
Merge conflict after #439/#440 landed — it's the anticipated trivial overlap on the commander import line in src/index.ts. Please rebase; everything else is approved and this merges immediately after. Reminder for the rebase: #393 will add 'x402' to the choices list on its side, so no action needed here for that.

ghozzza:
Rebased at `1ca1d88`. The overlap was as described: #439 added the `createRequire` version read and this PR added `Option` to the commander import, so the resolution keeps both.

Verified on the merged tree rather than assuming:

```
$ node dist/index.js --version
2.4.13                       # matches package.json, so #439 survived

$ node dist/index.js create x -t bogus
error: option '-t, --template <type>' argument 'bogus' is invalid.
Allowed choices are basic, farcaster-miniapp, minipay, ai-chat.
```

And `-t minipay -y --skip-install` still scaffolds `apps/contracts` and `apps/web`.

`x402` is deliberately absent from that list, per your note — #393 adds it on its side.

GigaHierz:
Rebased onto main as a maintainer edit — the only conflict was the anticipated commander/createRequire import line in src/index.ts; resolution is the union of both imports. Re-verified after rebase: CLI builds, `-t bogus` is rejected with the allowed-choices message, and --version still reports the real version.

Issue #421, opened 2026-08-11
Troubleshooting doc references CELO_PROVIDER_URL, which no template reads (actual: CELO_RPC_URL / CELO_SEPOLIA_RPC_URL)


GigaHierz:
## What happens

`docs/development/troubleshooting.mdx:64` tells users:

> **Check `.env` file**: Ensure your `PRIVATE_KEY` and `CELO_PROVIDER_URL` are correctly set up in your `.env` file.

`CELO_PROVIDER_URL` appears nowhere else in the repository. The variables the templates actually use are `CELO_RPC_URL` and `CELO_SEPOLIA_RPC_URL` (`templates/contracts/hardhat/.env.example.hbs:15-16`, consumed in `hardhat.config.ts.hbs`). A user following the troubleshooting step sets a variable nothing reads.

## Fix

Replace `CELO_PROVIDER_URL` with `CELO_RPC_URL` / `CELO_SEPOLIA_RPC_URL`.

Found in @celo/celo-composer 2.4.13 (main @ 710dd89).

ghozzza:
Fixed by #443, which merged on 12 Aug. It stayed open only because that PR's body carried a single `Closes #416` keyword — GitHub needs one per issue, so the other four it resolved were never linked.

Re-checked against current `main` (`f0750b5`) rather than assuming, since a merged PR is not the same as a verified fix:

```
$ grep -rn "CELO_PROVIDER_URL" docs/ README.md templates/
(no matches)
```

the CELO_PROVIDER_URL reference is gone from the repo. Safe to close.

GigaHierz:
Fixed by #443 — the CELO_PROVIDER_URL reference was dropped; the troubleshooting page now accurately says the RPC endpoints are written into hardhat.config.ts. The related dead-env-var cleanup is tracked in #451.

Issue #420, opened 2026-08-11
Docs claim Next.js build output is apps/web/dist — actual is .next; documented Vercel/Netlify settings break deploys


GigaHierz:
## What happens

The docs claim Next.js build output lands in `dist`, and tell users to configure deploy platforms accordingly:

- `docs/development/workflow.mdx:38` — "The output will be placed in the `dist` directory for each app (e.g., `apps/web/dist`)."
- `docs/development/deployment.mdx:36` — "**Output Directory**: `apps/web/dist` (or as configured in your `next.config.js`)."

`templates/base/apps/web/package.json.hbs:6` runs plain `next build`, and `templates/base/apps/web/next.config.js.hbs` sets no `distDir`/`output` — so the build output is `apps/web/.next`. A Vercel/Netlify project configured per the deployment doc points at a directory that doesn't exist and the deploy fails.

## Fix

Correct both references to `.next` (for Vercel, the output directory should normally be left at the framework default anyway).

Found in @celo/celo-composer 2.4.13 (main @ 710dd89).

ghozzza:
Fixed by #443, which merged on 12 Aug. It stayed open only because that PR's body carried a single `Closes #416` keyword — GitHub needs one per issue, so the other four it resolved were never linked.

Re-checked against current `main` (`f0750b5`) rather than assuming, since a merged PR is not the same as a verified fix:

```
$ grep -rn "apps/web/dist" docs/ README.md
(no matches)
```

the apps/web/dist build-output claim is gone from the repo. Safe to close.

GigaHierz:
Fixed by #443 — build output references corrected to .next, and the deployment doc now says to leave the output directory at the framework default.

Issue #419, opened 2026-08-11
Foundry docs reference a nonexistent .env.example and show deploy commands that cannot run (empty flags, unknown mainnet RPC alias)


GigaHierz:
## What happens

`docs/integrations/contracts/foundry.mdx` describes files and commands that don't exist or can't run:

1. **Nonexistent env file.** Line 30: "**`.env.example`**: A template for your environment variables, including `PRIVATE_KEY` and `ETHERSCAN_API_KEY`." — `templates/contracts/foundry/` contains only `foundry.toml.hbs`, `src/`, `script/`, `test/`, `.gitignore.hbs`. No `.env.example` is generated, and `Counter.s.sol.hbs` never reads a private key (`vm.envUint` absent), so line 50's instruction to set `PRIVATE_KEY` in "a `.env` file" leads nowhere.

2. **Malformed deploy command.** Line 54:
   ```
   forge script script/Counter.s.sol:CounterScript --rpc-url --chain-id --broadcast --verify
   ```
   `--rpc-url` and `--chain-id` are given with no values — this command cannot run as printed.

3. **Wrong RPC alias.** Line 57 uses `--rpc-url mainnet`, but `templates/contracts/foundry/foundry.toml.hbs:8-10` defines the aliases `celo-sepolia` and `celo`. There is no `mainnet` endpoint; the command fails with an unknown alias.

4. Line 20/34 also point at `packages/contracts`, which doesn't exist (covered in the contracts-directory issue).

## Fix

Rewrite the deploy section around the actual aliases (`--rpc-url celo-sepolia` / `--rpc-url celo`), with a complete working command, and either scaffold the promised `.env.example` or drop the claim (see also the companion issue about the foundry template being non-functional).

Found in @celo/celo-composer 2.4.13 (main @ 710dd89).

Issue #418, opened 2026-08-11
Docs instruct cp .env.example at project root, but the generated file is apps/web/.env.template


GigaHierz:
## What happens

Two docs instruct users to copy a root `.env.example` that is never generated:

- `docs/cli-reference/configuration.mdx:11` — "a `.env.example` file is generated in the root of your project", followed at line 18 by `cp .env.example .env`.
- `templates/farcaster-miniapp/FARCASTER_SETUP.md.hbs:30` (shipped into every farcaster project) — `cp .env.example .env.local`.

Reality: the base template's only web env file is `templates/base/apps/web/.env.template.hbs`, which renders to **`apps/web/.env.template`**. There is no root env file, and no file named `.env.example` outside `apps/contracts/` (hardhat only). Both documented `cp` commands fail with "No such file or directory" on a fresh scaffold.

## Fix

Update both docs to `cp apps/web/.env.template apps/web/.env.local` (or rename the generated file to `.env.example` for consistency with the contracts package — either way, make the docs and the generator agree).

Found in @celo/celo-composer 2.4.13 (main @ 710dd89).

ghozzza:
Fixed by #443, which merged on 12 Aug. It stayed open only because that PR's body carried a single `Closes #416` keyword — GitHub needs one per issue, so the other four it resolved were never linked.

Re-checked against current `main` (`f0750b5`) rather than assuming, since a merged PR is not the same as a verified fix:

```
$ grep -rn "cp .env.example" docs/ README.md
(no matches)
```

the root .env.example instruction is gone from the repo. Safe to close.

GigaHierz:
Fixed by #443 — configuration.mdx now documents the real env-file layout (apps/web/.env.template, apps/contracts/.env.example for hardhat) and the cp commands reference files that exist. FARCASTER_SETUP.md was corrected in the same PR.

PR #443, opened 2026-08-12
docs: name the directories, files and variables that actually exist

### LINKED ISSUE #416: Docs name the contracts directory wrongly everywhere: apps/hardhat and packages/contracts vs actual apps/contracts
## What happens

Contracts are generated at `apps/contracts` (`src/plopfile.ts:157,175` — destination `{{projectPath}}/apps/contracts/` for both hardhat and foundry), but the documentation names that directory wrongly almost everywhere — and inconsistently between pages.

| Location | Says | Actual |
|---|---|---|
| `docs/integrations/contracts/hardhat.mdx:11` | "sets up a dedicated `apps/hardhat` application" | `apps/contracts` |
| `docs/integrations/contracts/hardhat.mdx:23` | "In the `apps/hardhat` directory, you'll find a `.env.example`" | `apps/contracts/.env.example` |
| `docs/integrations/contracts/hardhat.mdx:45,46,51` | `apps/hardhat/contracts`, `apps/hardhat/test`, "`apps/hardhat` application" | `apps/contracts/...` |
| `docs/development/workflow.mdx:25` | "contracts in `apps/hardhat/contracts`" | `apps/contracts/contracts` |
| `docs/core-concepts/project-structure.mdx:14,27` | `apps/hardhat/` in tree + section | `apps/contracts/` |
| `docs/getting-started/quickstart.mdx:46` | "**`apps/hardhat`**" | `apps/contracts` |
| `docs/integrations/contracts/foundry.mdx:20` | "creates a new directory at **`packages/contracts`**" | `apps/contracts` |
| `docs/integrations/contracts/foundry.mdx:34` | "run from the `packages/contracts` directory" | `apps/contracts` |
| `templates/base/README.md.hbs:26` (generated project README) | "`apps/hardhat` - Smart contract development environment" | `apps/contracts` |

Note the docs also contradict *each other*: hardhat.mdx says `apps/hardhat`, foundry.mdx says `packages/contracts`, and neither matches the generator. Every `cd` command a new user copies from these pages fails.

## Fix

Replace all of the above with `apps/contracts`.

ghozzza:
Closes #416, #417, #418, #420, #421.

**Grouped deliberately.** These are five instances of one defect — documentation describing a project the CLI does not generate — and the diff is one pass over `docs/` plus two template files. Five separate PRs would be five reviews of the same shape. Happy to split if you would rather review them apart.

**#419 is deliberately excluded.** It is about the foundry docs referencing a `.env.example` that does not exist — and **#430 adds that file**. Fixing the docs against today's reality would make them wrong the moment that lands, so it waits.

## #416 — the contracts directory

Contracts generate at `apps/contracts` for **both** frameworks (`plopfile.ts:157,175`). The docs said `apps/hardhat` in five places and `packages/contracts` in two, so they were wrong *and* disagreed with each other.

**One instance the issue did not list, and it is the one users actually read:** `templates/base/README.md.hbs` — the README shipped into every generated project — also said `apps/hardhat`. Worse, its guard was `{{#if (eq contractFramework "hardhat")}}`, so a **foundry** project's README never mentioned its contracts directory at all. Now `(or hardhat foundry)`, verified:

```
-c hardhat  → apps/contracts
-c foundry  → apps/contracts
-c none     → (no contracts line, correctly)
```

## #417 — `packages/` does not exist

The README's structure block listed `packages/ui` and `packages/utils`. There is no `packages/` anywhere in `templates/`, and `pnpm-workspace.yaml.hbs` declares only `apps/*`. The generated project's own README already says the opposite — "embedded UI components and utilities". Block removed.

## #418 — the root `.env.example`

Two places told users to copy a file that is never generated. The web env file is `apps/web/.env.template`, so both now name it — verified present in a generated project.

**One correction to the issue:** `templates/contracts/hardhat/README.md.hbs:52` also says `cp .env.example .env`, and that one is **right** — the hardhat template does ship `.env.example.hbs`. Left alone.

## #420 — `apps/web/dist`

`next build` with no `distDir` writes `.next`. `dist` has never existed.

The deployment doc is the harmful one: it told people to enter `apps/web/dist` as the platform's **Output Directory**, which breaks a deploy rather than merely misleading. Replaced with "leave blank" — Vercel and Netlify detect Next.js and find `.next` themselves.

## #421 — `CELO_PROVIDER_URL`

Appears nowhere else in the repository. The templates read `CELO_RPC_URL` and `CELO_SEPOLIA_RPC_URL`, and both have defaults in `hardhat.config.ts`, so the troubleshooting step now says to set them only if you overrode the defaults — and points at `apps/contracts/.env`, which is where they belong.

## Verified

Every claim checked against the source rather than taken from the issue, and a sweep afterwards confirms no `apps/hardhat`, `packages/contracts`, `packages/ui`, `apps/web/dist` or `CELO_PROVIDER_URL` remains anywhere in `docs/`, `README.md` or `templates/`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,9 +153,6 @@ my-celo-app/
 ├── apps/
 │   ├── web/                 # Next.js application
 │   └── contracts/           # Smart contracts (if selected)
-├── packages/
-│   ├── ui/                  # Shared UI components
-│   └── utils/               # Shared utilities
 ├── package.json             # Root package.json
 ├── pnpm-workspace.yaml      # PNPM workspace config
 ├── turbo.json              # Turborepo configuration
--- a/docs/cli-reference/configuration.mdx
+++ b/docs/cli-reference/configuration.mdx
@@ -8,14 +8,20 @@ Your Celo Composer project is configured primarily through environment variables
 
 ## Environment Variables
 
-When you create a new project, a `.env.example` file is generated in the root of your project. To set up your local environment, you should:
+Nothing is generated at the root. Each app carries its own template, and the file is named differently in each:
 
-1.  **Create a copy** of this file and name it `.env`.
-2.  **Fill in the required values** in the `.env` file.
+| app | template file | copy it to |
+|---|---|---|
+| `apps/web` | `.env.template` | `apps/web/.env` |
+| `apps/contracts` (Hardhat only) | `.env.example` | `apps/contracts/.env` |
+
+To set up your local environment, copy each template that your project generated and fill in the values:
 
 ```bash
-# Copy the example file
 cp apps/web/.env.template apps/web/.env
+
+# only if you chose Hardhat — a Foundry project generates no env template
+cp apps/contracts/.env.example apps/contracts/.env
 ```
 
 <Warning>
--- a/docs/cli-reference/configuration.mdx
+++ b/docs/cli-reference/configuration.mdx
@@ -15,7 +15,7 @@ When you create a new project, a `.env.example` file is generated in the root of
 
 ```bash
 # Copy the example file
-cp .env.example .env
+cp apps/web/.env.template apps/web/.env
 ```
 
 <Warning>
--- a/docs/core-concepts/project-structure.mdx
+++ b/docs/core-concepts/project-structure.mdx
@@ -24,7 +24,7 @@ The `apps` directory contains the individual applications within your monorepo. 
 
 - **`apps/web`**: This is the main Next.js web application that serves as the user-facing frontend of your dApp.
 
-- **`apps/hardhat`** (Optional): If you choose to include the Hardhat framework, it will be set up as a separate application here. It contains your Solidity smart contracts, deployment scripts, and tests.
+- **`apps/contracts`** (Optional): If you choose to include the Hardhat framework, it will be set up as a separate application here. It contains your Solidity smart contracts, deployment scripts, and tests.
 
 ## Tooling
 
--- a/docs/core-concepts/project-structure.mdx
+++ b/docs/core-concepts/project-structure.mdx
@@ -11,11 +11,12 @@ Here's a typical directory layout:
 ```plaintext
 my-celo-app/
 ├── apps/
-│   ├── web/          # Next.js frontend application
-│   └── hardhat/      # Optional: Hardhat smart contracts
-├── package.json      # Root package.json
+│   ├── web/            # Next.js frontend application
+│   └── contracts/      # Optional: Hardhat or Foundry smart contracts
+├── package.json        # Root package.json
 ├── pnpm-workspace.yaml # Defines the workspace
-└── turbo.json        # Turborepo configuration
+├── tsconfig.json       # Base TypeScript config
+└── turbo.json          # Turborepo configuration
 ```
 
 ## `apps` Directory
--- a/docs/development/deployment.mdx
+++ b/docs/development/deployment.mdx
@@ -33,7 +33,7 @@ While the specific steps may vary depending on the platform you choose, the gene
 
     - **Framework Preset**: Select `Next.js`.
     - **Build Command**: `pnpm build`.
-    - **Output Directory**: `apps/web/dist` (or as configured in your `next.config.js`).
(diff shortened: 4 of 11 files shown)


GigaHierz (changes requested):
Most of this PR is verified correct (all the hardhat.mdx / workflow.mdx / quickstart.mdx path fixes, the README `packages/` removal, the `.env.template` cp commands, the `.next` build-output fixes, and the base README `or`-helper change all check out). Three items from the issue set remain unfixed in the current head:

1. **`docs/core-concepts/project-structure.mdx` directory tree (lines 11–18).** The prose was fixed but the tree still shows `└── hardhat/  # Optional: Hardhat smart contracts`. The page now contradicts itself — prose says `apps/contracts`, diagram says `hardhat/`. (Issue #416 lists both line 14 and line 27.)

2. **`docs/cli-reference/configuration.mdx` prose (lines 11–13).** Still says "a `.env.example` file is generated in the root of your project… Create a copy of this file and name it `.env`" directly above the corrected `cp apps/web/.env.template` command. The prose and numbered steps need the same fix the command got. (Issue #418 lists both the prose and the command.)

3. **`docs/development/troubleshooting.mdx:64` now makes a stronger false claim.** The amended text says "if you overrode the defaults, `CELO_RPC_URL` / `CELO_SEPOLIA_RPC_URL` are set in `apps/contracts/.env`" — but `templates/contracts/hardhat/hardhat.config.ts.hbs` (lines 17 and 23) hardcodes both RPC URLs and never reads those env vars. There is no override mechanism; the vars exist only in `.env.example` and are dead. Either drop the RPC sentence entirely, or (better, as a separate template fix) make the hardhat config actually read them with the forno URLs as fallbacks — then this sentence becomes true.

ghozzza:
All three at `f1c538d`, each checked against a generated project rather than against the page above it.

**1. The structure diagram.** `apps/hardhat/` → `apps/contracts/`, so it stops contradicting its own prose two paragraphs down. Two things the diff also picked up from the real scaffold: the optional app is *"Hardhat or Foundry smart contracts"* — a Foundry project lands in the same directory — and the root `tsconfig.json` was missing from the tree entirely.

**2. The configuration prose.** You were right that fixing the command and leaving the prose was half a fix. But the prose was wrong in a second way too: there is no single env file. A generated project has **two, in different places, with different names**, and which ones exist depends on the contract framework:

| `create …` | files generated |
|---|---|
| `-c hardhat` | `apps/web/.env.template` **and** `apps/contracts/.env.example` |
| `-c foundry` | `apps/web/.env.template` only |
| `-c none` | `apps/web/.env.template` only |

The section is now a table plus the two `cp` lines, with the Foundry case called out — the missing Foundry env file is #419's, not fixed here.

**3. The RPC claim — took the first of your two options, and filed the second.** The sentence is gone; the step now says `PRIVATE_KEY` is the only variable that config reads, and that the endpoints live in `hardhat.config.ts`.

Confirmed the underlying finding first, and it is worse than the one line suggested:

```
$ grep -rn "CELO_RPC_URL\|CELO_SEPOLIA_RPC_URL" templates/ src/
templates/contracts/hardhat/.env.example.hbs:15
templates/contracts/hardhat/.env.example.hbs:16
templates/base/apps/web/.env.template.hbs:3
```

Three declarations, zero readers. `apps/web/.env.template` ships a dead `CELO_RPC_URL` too, which the issue did not mention.

The template fix you preferred is **#451**, with the `?? "https://forno…"` fallback form so defaults do not change, a note that #427 should land first since it edits the same file, and the open question of what the web app's copy is for. Kept out of this PR because it is a template change and this one is docs.

GigaHierz (approved):
Re-verified at f1c538d: tree diagram fixed, configuration.mdx now documents the real env-file layout (verified against the templates), and the troubleshooting RPC sentence was dropped in favor of an accurate description. Two wording nits, non-blocking: 'the only variable the Hardhat config reads' overstates (it also reads ETHERSCAN_API_KEY), and the project-structure bullet still says 'the Hardhat framework' while the tree says 'Hardhat or Foundry'. Pre-existing and out of scope, but worth a follow-up issue: the hardhat .env.example lists CELO_RPC_URL/CELO_SEPOLIA_RPC_URL that nothing reads, and the template has no dotenv, so apps/contracts/.env isn't auto-loaded.

Issue #423, opened 2026-08-11
ai-chat README describes features that are stubbed out: NextAuth auth, wrong structure diagram, unwired provider env vars


GigaHierz:
## What happens

The README shipped inside every `ai-chat` scaffold (`templates/ai/chat-template/README.md`) materially misdescribes the code it sits next to:

1. **Auth doesn't exist.** The README claims "**User Authentication**: Secure login and guest access" and lists "Authentication: NextAuth.js" in the tech stack (lines 11, 84). But `app/(auth)/auth.ts` is entirely stubbed — first line: `// Auth disabled; export stubs for compatibility`, returning a hardcoded `{ id: "public-user" }` — `middleware.ts` allows all requests ("public mode"), and `next-auth` is not in `package.json` at all. Meanwhile the dead surface still ships: `app/(auth)/api/auth/[...nextauth]/route.ts`, `login/page.tsx`, `register/page.tsx`, and an `AUTH_SECRET` entry in `.env.example` that nothing reads.

2. **Wrong structure diagram.** The "Project Structure" section shows `apps/web/app/…`, but the template is flat (`app/`, `components/`, `lib/` at the root — there is no `apps/` directory in an ai-chat scaffold).

3. **Env vars for providers that aren't wired.** `.env.example` and the README advertise `DEEPSEEK_API_KEY` and `XAI_API_KEY`, but `lib/ai/providers.ts` wires only Google, OpenAI, Anthropic, and Mistral — there is no deepseek or xai provider package in `package.json`. Setting those keys does nothing.

## Fix

Either remove the stubbed auth surface + claims (and the dead env vars), or restore real auth — but make the README describe the template that actually ships. Fix the structure diagram either way.

(Related: #387 covers ai-chat scaffolding/deployability; this issue is specifically about the shipped README contradicting the shipped code.)

Found in @celo/celo-composer 2.4.13 (main @ 710dd89).

GigaHierz:
Correction to point 3 of this issue: `DEEPSEEK_API_KEY` and `XAI_API_KEY` **are** wired. `lib/ai/providers.ts:128-168` reads both keys and registers models through `createOpenAI({ apiKey, baseURL })` against the DeepSeek and x.ai OpenAI-compatible endpoints — no dedicated `@ai-sdk/deepseek`/`@ai-sdk/xai` package is needed, and `@ai-sdk/openai` is a declared dependency. Setting either key adds working models to the picker.

[...]

PR #444, opened 2026-08-12
docs: fix the assorted inaccuracies, and give ai-chat a page

### LINKED ISSUE #425: Assorted docs inaccuracies: non-Mintlify Callout, duplicate H1, unexported USDT_ADDRESS reuse claim, missing ai-chat docs page, shadcn claim
## Summary

Grouped small documentation inaccuracies — none blocking on its own, all cheap to fix:

1. **Non-Mintlify component.** `docs/templates/minipay.mdx:12,17` uses `<Callout type="info">`; Mintlify's components are `<Note>/<Info>/<Warning>/<Tip>/<Check>` (every other page uses `<Info>`/`<Warning>` correctly), so these blocks won't render as intended.

2. **Duplicate H1.** `docs/templates/minipay.mdx` has both frontmatter `title: Minipay Template` and a literal `# Minipay Template` at line 8 — Mintlify renders the title, producing a doubled heading. No other page does this.

3. **Wrong reuse instruction.** `docs/templates/minipay.mdx:115` says the `UserBalance` component "already imports `USDT_ADDRESS` — use the same constant", but `templates/minipay/components/user-balance.tsx.hbs:8` declares it as a module-local, unexported `const`; it can't be imported as written.

4. **ai-chat has no docs page.** `docs/docs.json` navigation lists only `templates/web-app-template`, `templates/farcaster-miniapp`, `templates/minipay`. The `ai-chat` template is mentioned in the README and CLI reference but is unreachable from docs navigation.

5. **shadcn/ui "pre-configured" claim.** `README.md:9` and `docs/templates/web-app-template.mdx:18` — the base template ships three hand-copied components (`ui/button`, `ui/card`, `ui/sheet`) but no `components.json`, so `npx shadcn add …` doesn't work out of the box in a generated project.

6. **Install-path inconsistency.** `docs/getting-started/installation.mdx` teaches npx-only, while `docs/development/troubleshooting.mdx:37-42` troubleshoots a global-install workflow (`npm install -g @celo/celo-composer`) the docs never introduce.

7. **Stale link.** `docs/integrations/wallets/rainbowkit.mdx:19` points at `cloud.walletconnect.com`, which now redirects to Reown Cloud.

8. **`--force` claim.** `docs/development/troubleshooting.mdx:58` presents `pnpm contracts:compile --force` as a "clean compilation"; the flag busts the Turborepo cache, not Hardhat's — the equivalent is `hardhat clean && hardhat compile`.

9. **"Next.js 14+" everywhere.** `README.md:7,76,100,181`, `docs/getting-started/introduction.mdx:28`, `docs/templates/web-app-template.mdx:14` — true for the base template but the ai-chat template ships Next 15.3.6; worth phrasing per-template.

ghozzza:
Closes #425 — eight of the nine items. Item 3 is held back, reasoning at the bottom.

| # | | |
|---|---|---|
| 1 | `<Callout>` is not a Mintlify component | → `<Info>`, which every other page already uses |
| 2 | literal `# Minipay Template` under a frontmatter `title` | removed the duplicate heading |
| 4 | ai-chat unreachable from navigation | **new page** + nav entry |
| 5 | "Pre-configured shadcn/ui components" | three components are copied in and there is no `components.json`, so `shadcn add` needs `shadcn init` first — says so now |
| 6 | troubleshooting answered for a global install | the docs teach `npx`, which needs no install |
| 7 | `cloud.walletconnect.com` | → `cloud.reown.com` |
| 8 | `contracts:compile --force` as "clean compilation" | busts the **Turborepo** cache, not Hardhat's — replaced with `hardhat clean` |
| 9 | "Next.js 14+" everywhere | true of the base templates, not of ai-chat at 15.3.6 — stated per template |

## The ai-chat page

Written from the template rather than from its README, because the README overclaims — which is #423, and worth not repeating here:

- **It is not a monorepo.** No `apps/web`, no `apps/contracts`, no Turborepo. `--wallet-provider` and `--contracts` do not apply.
- **`POSTGRES_URL` is read at build time**, so `pnpm build` fails without a database rather than starting and erroring later. That is the first thing anyone hits.
- **Authentication is stubbed.** `app/(auth)/auth.ts` opens with `// Auth disabled; export stubs for compatibility` and returns a fixed `public-user`; `middleware.ts` allows every request; `next-auth` is not a dependency. The page says so plainly, and that every visitor therefore shares one identity.
- Four providers are actually wired through the AI SDK — Anthropic, Google, Mistral, OpenAI — verified from `lib/ai/providers.ts` rather than from the feature list.

`docs.json` still parses after the nav edit.

## One thing I corrected beyond the issue

While in `docs/templates/minipay.mdx`, the token table listed `0x765DE816845861e75A25fCA122bb6898B8B1282a` as **cUSD**. Read from the contract on mainnet, it answers `name() = "Mento Dollar"`, `symbol() = "USDm"` — same token, renamed on chain. The table now says so.

## Item 3, held back

It says `docs/templates/minipay.mdx:115` wrongly tells readers to reuse `USDT_ADDRESS` from `UserBalance`, which is module-local and unexported. **#432 replaces that constant with an exported `TOKENS_BY_CHAIN`**, so the correct instruction depends on which of the two lands first. Fixing it now would make the doc wrong again the moment that PR merges. Same reason #419 is not in #443.

## One self-correction worth flagging

My first pass at item 8 wrote `pnpm --filter contracts exec hardhat clean`. The package is named **`hardhat`**, not `contracts` — `templates/contracts/hardhat/package.json.hbs:2`. Caught before pushing; the command in the diff is the working one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ A powerful CLI tool for generating customizable Celo blockchain starter kits wit
 
 ## Features
 
-- 🚀 **Modern Stack**: Next.js 14+, TypeScript, Tailwind CSS, shadcn/ui
+- 🚀 **Modern Stack**: Next.js (14+ in the base templates, 15 in ai-chat), TypeScript, Tailwind CSS, shadcn/ui
 - 📦 **Monorepo Ready**: Turborepo with PNPM workspaces
-- 🎨 **Beautiful UI**: Pre-configured shadcn/ui components
+- 🎨 **Beautiful UI**: shadcn/ui components copied in (button, card, sheet) — add more with the shadcn CLI after running `npx shadcn init`
 - 🔧 **Developer Experience**: Interactive prompts and clear feedback
 - 🌍 **Celo Optimized**: Ready for Celo blockchain development
 - 🎯 **Multiple Templates**: Choose from Basic Web App, Farcaster Miniapp, Minipay, or AI Chat templates
--- a/docs/development/troubleshooting.mdx
+++ b/docs/development/troubleshooting.mdx
@@ -36,10 +36,9 @@ export default function Page() {
 
 **Problem**: `celo-composer: command not found`
 
-**Solution**: This usually means the global installation didn't work correctly or your shell's `PATH` is not configured to find global Node modules.
+**Solution**: The docs use `npx`, which needs no installation at all — run `npx @celo/celo-composer@latest create` and the command is fetched on demand.
 
-1.  **Verify Installation**: Make sure you have installed the package globally with `npm install -g @celo/celo-composer`.
-2.  **Check PATH**: Ensure that your Node.js global bin directory is in your shell's `PATH`. You can find the directory by running `npm config get prefix`.
+If you installed it globally on purpose, check that your Node global bin directory is on your `PATH`. `npm config get prefix` prints where that is.
 
 **Problem**: Errors during `pnpm install`.
 
@@ -55,7 +54,7 @@ export default function Page() {
 **Solution**:
 
 1.  **Check Solidity Version**: Ensure the `solidity` version in your `hardhat.config.ts` matches the pragma statement in your contract files.
-2.  **Clean Cache**: Run `pnpm contracts:compile --force` to do a clean compilation.
+2.  **Clean Cache**: Run `pnpm --filter hardhat exec hardhat clean`, then `pnpm contracts:compile`. `--force` busts the Turborepo cache rather than Hardhat's, so it does not clear stale artifacts.
 
 **Problem**: Deployment script fails due to authentication or network issues.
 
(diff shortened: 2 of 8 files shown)


ghozzza:
Fixed at `13b0fd0`. The page now says six, and the env table carries `DEEPSEEK_API_KEY` and `XAI_API_KEY` with the models each one unlocks.

Added one clause explaining *why* they need no extra package, since that is what made the count look like four in the first place: both reach their provider through `@ai-sdk/openai` against an OpenAI-compatible endpoint, not through a dedicated `@ai-sdk/*` package. A dependency check alone will keep producing the wrong answer here.

Swept the rest of the PR for the same claim — no other page counts providers.

GigaHierz (approved):
Verified: six providers now documented (matches providers.ts:128-168), env table complete, ai-chat page registered in docs.json and factually accurate against the template, USDm relabel verified on-chain, filter commands correct. Includes the MiniPay token-table update pairing with #432. Nit: stray indentation on the docs.json entry — cosmetic.

PR #445, opened 2026-08-12
docs(templates): make the ai-chat README describe the code beside it

### LINKED ISSUE #423: ai-chat README describes features that are stubbed out: NextAuth auth, wrong structure diagram, unwired provider env vars

ghozzza:
Closes #423.

The README ships **inside every ai-chat scaffold**, so it is the first thing a user reads about code they now own.

## 1. Authentication was claimed and does not exist

The README advertised "**User Authentication**: Secure login and guest access" and listed "Authentication: NextAuth.js" in the tech stack. In the same tree:

```
app/(auth)/auth.ts:1   // Auth disabled; export stubs for compatibility
                       returns a hardcoded { id: "public-user" }
middleware.ts          allows all requests
package.json           next-auth is not a dependency
```

This is the worst of the three, because the failure mode is silent: a reader assumes their deployment is protected. The README now has an **Authentication** section stating plainly that every visitor shares one identity, and that the `(auth)` route group is scaffolding to build on rather than a working flow.

## 2. The structure diagram showed a monorepo

It drew `apps/web/…`. An ai-chat scaffold is flat — `app/`, `components/`, `lib/` at the root, no `apps/` directory at all. Corrected, and the `(auth)` line is annotated as stubbed.

## 3. Two provider keys that nothing reads

`DEEPSEEK_API_KEY` and `XAI_API_KEY` were in both the README and `.env.example`. `lib/ai/providers.ts` wires Google, OpenAI, Anthropic and Mistral, and neither provider package is in `package.json` — setting those keys did nothing. Removed.

`AUTH_SECRET` stays in `.env.example` but commented out, with a line saying nothing reads it, so it does not silently reappear as a thing to configure.

## Deliberately not included

**The dead auth surface itself** — `login/page.tsx`, `register/page.tsx`, `app/(auth)/api/auth/[...nextauth]/route.ts`. The issue offers "remove the stubbed surface *or* restore real auth"; both are code changes with their own blast radius, and this issue is specifically the README contradicting the code. The README now points at those files as scaffolding, which is true either way. Happy to take the removal, or a real auth wiring, as a follow-up.

Related: the docs-site page for this template (added in #444) says the same things, so the two do not drift apart again.

### DIFF
--- a/templates/ai/chat-template/.env.example
+++ b/templates/ai/chat-template/.env.example
@@ -1,5 +1,6 @@
 # Generate a random secret: https://generate-secret.vercel.app/32 or `openssl rand -base64 32`
-AUTH_SECRET=
+# Nothing reads this. The (auth) route group is stubbed — see the README.
+# AUTH_SECRET=
 
 NEXT_PUBLIC_THIRDWEB_CLIENT_ID=
 
@@ -18,12 +19,6 @@ ANTHROPIC_API_KEY=
 
 # Mistral: https://console.mistral.ai/api-keys/
 MISTRAL_API_KEY=
-
-# DeepSeek (OpenAI-compatible): https://platform.deepseek.com/api-keys
-DEEPSEEK_API_KEY=
-
-# xAI (OpenAI-compatible): https://console.x.ai/
-XAI_API_KEY=
 
 # Google Gemini: https://aistudio.google.com/app/apikey
 GOOGLE_GENERATIVE_AI_API_KEY=
--- a/templates/ai/chat-template/.env.example
+++ b/templates/ai/chat-template/.env.example
@@ -20,6 +20,12 @@ ANTHROPIC_API_KEY=
 # Mistral: https://console.mistral.ai/api-keys/
 MISTRAL_API_KEY=
 
+# DeepSeek (OpenAI-compatible): https://platform.deepseek.com/api-keys
+DEEPSEEK_API_KEY=
+
+# xAI (OpenAI-compatible): https://console.x.ai/
+XAI_API_KEY=
+
 # Google Gemini: https://aistudio.google.com/app/apikey
 GOOGLE_GENERATIVE_AI_API_KEY=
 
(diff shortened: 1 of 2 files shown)


ghozzza:
Restored at `70a2f03`. Checked the code before putting them back rather than taking the correction on trust:

`lib/ai/providers.ts:128-165` gates each on its own key and registers it through `createOpenAI` with a custom `baseURL` — `https://api.deepseek.com` and `https://api.x.ai/v1`. Four models between them: `deepseek-chat`, `deepseek-reasoner`, `grok-2-mini`, `grok-2-latest`. So `@ai-sdk/openai` being the only declared package is the design, not a gap.

`.env.example` now differs from `main` in exactly one place — the `AUTH_SECRET` line, which is point 1 of #423 and stands.

The README needed no change: line 7 already names all six providers on `main`, and this branch never touched it.

GigaHierz (approved):
Verified: DeepSeek/xAI env keys restored (both are wired via OpenAI-compatible endpoints), net env change is now only the AUTH_SECRET comment-out, and the auth-stub description matches app/(auth)/auth.ts exactly.

PR #446, opened 2026-08-12
docs: catch the changelog up from 2.3.5 to today

### LINKED ISSUE #422: CHANGELOG stops at 2.3.5 while 2.4.13 is published — ~8 releases undocumented, Keep-a-Changelog structure broken
## What happens

`CHANGELOG.md`'s latest entry is `## [2.3.5] - 2025-08-06`, while the published package is **2.4.13**. Git tags confirm at least `v2.3.6`, `v2.3.7`, `v2.4.1`, `v2.4.2`, `v2.4.3`, `v2.4.4`, `v2.4.6`, `v2.4.10`, plus the 2.4.13 bump — roughly eight releases with no changelog entry.

The undocumented window includes the largest user-facing changes in the tree:

- the Alfajores → **Celo Sepolia** migration
- Foundry contract-framework support
- the `ai-chat` template
- the chat-template Next.js bump for GHSA-9qr9-h5gf-34mp

Structural problems in the file itself:

- The Keep-a-Changelog boilerplate ("All notable changes… Semantic Versioning") sits between the 2.3.3 and 2.2.4 entries instead of at the top.
- `[Unreleased]` is positioned at the bottom (line 88), below all released versions.
- Version ordering jumps 2.3.3 → 2.2.4 with nothing in between.
- Line 77 still says "Pre-configured for Celo mainnet and **Alfajores** testnet" — the only remaining Alfajores reference in the repo, now factually wrong.
- The install example at line 120 recommends pinning `@2.2.0`, which predates everything documented.

`scripts/release.sh:39-41` relies on the releaser interactively confirming a changelog update, which evidently isn't happening.

## Fix

Backfill entries for 2.3.6 → 2.4.13 (even one-liners), restore Keep-a-Changelog structure (`Unreleased` on top, boilerplate under the title), and fix the stale Alfajores line. Longer-term, generate entries in the release script instead of prompting.

ghozzza:
Closes #422.

The file stopped at `2.3.5` while the published package was `2.4.13`. The undocumented window contains the biggest user-facing changes in the repository — the **Alfajores → Celo Sepolia** migration, **Foundry** support, the **MiniPay** and **ai-chat** templates, the **ESM** migration, and the chat-template Next bump for GHSA-9qr9-h5gf-34mp.

## How the entries were derived

From the tags and `git log` per release window, not from memory. Some of what that turned up is worth stating rather than smoothing over:

- **`v2.3.5` is not a tag.** The changelog's most recent entry names a version that was never tagged.
- **2.4.7, 2.4.8 and 2.4.9 have no tags either.** Their bumps appear as commits inside the `v2.4.10` window, so they are grouped under 2.4.10 with a line saying so.
- **2.4.11 and 2.4.12 have no bump commits at all** — the next bump after 2.4.10 goes straight to 2.4.13.
- **Twelve commits sit after the 2.4.13 bump with no release.** They are under `## [Unreleased]` rather than attributed to a version that did not ship them.

Where the history does not support a split, the file says so instead of inventing one.

## The structural fix

The issue notes the Keep-a-Changelog structure is broken. It was worse than a missing header: the preamble —

> All notable changes to this project will be documented in this file.
> The format is based on Keep a Changelog…

— was sitting **below three version entries**, so the file opened in the middle of its own history. Moved to the top, with the Semantic Versioning line the format expects.

## Result

```
## [Unreleased]
## [2.4.13] - 2025-12-18
## [2.4.10] - 2025-08-21    ← covers 2.4.7–2.4.10, published untagged
## [2.4.6]  - 2025-08-21
## [2.4.4]  - 2025-08-21
## [2.4.3]  - 2025-08-18
## [2.4.2]  - 2025-08-18
## [2.4.1]  - 2025-08-18
## [2.3.7]  - 2025-08-07
## [2.3.6]  - 2025-08-07
## [2.3.5]  - 2025-08-06    ← previously the newest entry
```

Existing entries below 2.3.5 are untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,93 @@
 # Changelog
 
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+- Miniapp environment detection in the Farcaster template.
+
+### Fixed
+- Bumped `next` to 15.3.6 in the chat template for GHSA-9qr9-h5gf-34mp.
+- Wallet status display no longer shows a hardcoded address; authentication status message corrected.
+- Variable initialisation and SDK property access errors in the Farcaster template.
+- Removed duplicate keys and fixed formatting across the `.env.example` files (#372).
+
+### Documentation
+- MiniPay: token reference table, a USDT transfer example, and JSDoc on `UserBalance` covering addresses and decimals.
+
+## [2.4.13] - 2025-12-18
+
+### Changed
+- **Alfajores replaced by Celo Sepolia throughout** — templates, docs and network config.
+- Contract verification moved to the Etherscan V2 API in place of Celoscan.
+- thirdweb bumped to a version that includes `celoSepoliaTestnet`.
+
+### Fixed
+- Plopfile and template paths now resolve on Windows.
+- Connect-button generation.
+
+### Documentation
+- README rewritten around the full template set and the command options.
+
+## [2.4.10] - 2025-08-21
+
+Covers 2.4.7 through 2.4.10, which were published without tags.
+
+### Changed
+- **Codebase migrated to ESM modules**, with dependencies updated to match.
+- ESLint config converted to ESM, then to `.json`.
+
+## [2.4.6] - 2025-08-21
+
+### Fixed
+- Downgraded `plop` and `node-plop` to resolve version conflicts.
+
+## [2.4.4] - 2025-08-21
+
+### Fixed
+- Downgraded `chalk`, `inquirer` and `ora` to resolve ESM compatibility issues.
+
+## [2.4.3] - 2025-08-18
+
+### Fixed
+- `.env` added to the AI chat template's gitignore.
+
+## [2.4.2] - 2025-08-18
+
+### Documentation
+- AI chat template: Celo integration details and project structure.
+
+## [2.4.1] - 2025-08-18
+
+### Added
+- **AI chat template**, with document management and real-time streaming.
+- Jest and ts-jest configured for unit testing.
+
+## [2.3.7] - 2025-08-07
+
+### Added
+- **Foundry support** as a contract framework.
+- **MiniPay template**, with custom wallet integration and balance display.
+
+### Changed
+- Contracts moved to `apps/`, and the Git initialisation flow improved.
+
+### Documentation
+- MiniPay template documentation, and template references updated.
+
+## [2.3.6] - 2025-08-07
+
+### Fixed
+- Template generation conflict for `connect-button`.
+- `moduleResolution` set to `bundler` in the base template's tsconfig.
+
+### Documentation
+- Mintlify configuration updated; Farcaster template tag added.
+
 ## [2.3.5] - 2025-08-06
 
 ### Fixed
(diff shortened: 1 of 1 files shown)


GigaHierz (changes requested):
The backfilled content is excellent — every date and entry was independently verified against the git tags and commit windows, including the untagged-versions notes. What blocks this is purely file structure, and all of it was flagged before:

1. **Dangling boilerplate fragment.** The diff removes CHANGELOG lines 18–20 but leaves line 21 — "and this project adheres to [Semantic Versioning](…)." — stranded mid-file between the 2.3.3 and 2.2.4 entries, now duplicating the same sentence added at the top. Delete the leftover line.

2. **Duplicate `[Unreleased]`.** The old `## [Unreleased]` with its stale "Planned" list (~line 88 on main) is untouched, so the merged file has two `[Unreleased]` headings — one of the items issue #422 explicitly lists. Remove it or merge anything still relevant into the new top section.

3. **Stale Alfajores line (line 77).** "Pre-configured for Celo mainnet and Alfajores testnet" inside the historical 2.2.0 entry. Leaving history as-written is a defensible choice, but then say so — a footnote or a body note. Right now it's simply unaddressed while the issue lists it.

4. Nit: 2.4.5 was also published untagged (its bump commit precedes v2.4.6) — the 2.4.6 entry could carry a "covers 2.4.5–2.4.6" note like the one 2.4.10 has for 2.4.7–2.4.9.

ghozzza:
All four at `e25264a`.

**1. Stranded boilerplate — deleted.** The orphaned *"and this project adheres to Semantic Versioning"* line between 2.3.3 and 2.2.4 is gone. `grep -c "adheres to \[Semantic Versioning\]"` now returns **1**, at the header where it belongs.

**2. Duplicate `[Unreleased]` — removed.** `grep -c "^## \[Unreleased\]"` returns **1**. Nothing was merged upward: the old *Planned* list was five undated aspirations (*"Documentation improvements"*, *"Performance optimizations"*) sitting below a January 2024 entry. None of it describes work in flight, so carrying it into the live section would have been noise rather than history.

**3. Alfajores — kept, and now said out loud.** You called leaving history as-written defensible but unstated, which was the right read. There is now a policy line in the header:

> Released entries are left as they were written. Where a statement was true at the time and is not now — 2.2.0 describing Alfajores as the testnet, for instance — it stays, because a changelog records what shipped rather than what is currently true. The move to Celo Sepolia is its own entry under 2.4.13.

Stated once at the top rather than as a footnote per entry, so it also covers the next reader who spots a stale line further down.

**4. 2.4.5 — confirmed and noted.** Verified rather than taken on trust:

```
$ git tag -l 'v2.4.*'
v2.4.1 v2.4.2 v2.4.3 v2.4.4 v2.4.6 v2.4.10
```

No `v2.4.5`, and the 2.4.5 bump `a2e812b` is an ancestor of `v2.4.6`. The 2.4.6 entry now carries that, with the commit named so it is checkable.

---

**One addition you did not ask for, from #440's review.** You asked there for a release-note line covering the deliberate behavioural break, and that file is this PR's whole diff — putting it in #440 would collide here for no benefit. So it is in `[Unreleased] → Changed`, calling out that `create app -t basic --skip-install` used to run to completion in CI and now stops at the description prompt. Noted on #440 too so it cannot fall between the two.

GigaHierz (approved):
Re-verified at e25264a: single [Unreleased], no stranded boilerplate, Alfajores explicitly acknowledged in the preamble, 2.4.5 note verified against tags. Two conditions/nits: (1) MERGE ORDER — the [Unreleased] 'breaking for scripts' entry describes #440, so #440 must merge before this PR; (2) npm shows 2.4.11 and 2.4.12 published 2025-10-16 with no entry or coverage note — inconsistent with the PR's own convention under 2.4.6/2.4.10; and the 2.4.10 note's 'published without tags' wording is imprecise (v2.4.10 is tagged; 2.4.7–2.4.9 were never published). Fine as a fast-follow.

Issue #429, opened 2026-08-11
viem and wagmi are declared twice when farcaster-miniapp is combined with a wallet provider


ghozzza:
Found while fixing #400. Same shape as the `@tanstack/react-query` duplicate that #400 resolves, but for two other dependencies, so filing separately rather than widening that PR.

## What happens

```bash
npx @celo/celo-composer@2.4.13 create demo -t farcaster-miniapp --wallet-provider rainbowkit -c none --skip-install -y
```

The generated `apps/web/package.json` declares `viem` and `wagmi` **twice each**:

```json
    "@rainbow-me/rainbowkit": "^2.0.0",
    "viem": "^2.0.0",
    "wagmi": "^2.0.0",
    ...
    "viem": "^2.27.2",
    "wagmi": "^2.14.12",
```

It parses, because `JSON.parse` keeps the last occurrence — so the rainbowkit pins are silently discarded and the project resolves `viem@^2.27.2` / `wagmi@^2.14.12`. Nothing warns.

## Why the combination is reachable

`src/commands/create.ts:64` only defaults `walletProvider` to `none` for `farcaster-miniapp` **when the flag is absent**:

```ts
const autoWallet = options.walletProvider
  ? options.walletProvider
  : autoTemplateType === "farcaster-miniapp" || autoTemplateType === "ai-chat"
  ? "none"
  : ...
```

So passing `--wallet-provider rainbowkit` (or `thirdweb`) alongside `-t farcaster-miniapp` renders both the wallet branch and the farcaster branch of `templates/base/apps/web/package.json.hbs`, and the two lists overlap.

## Suggested fix

Same treatment #400 applies to `@tanstack/react-query` — declare each shared dependency once, guarded by an `or` over the branches that need it, and pick the stricter of the two ranges. `viem` and `wagmi` are the only remaining overlaps.

Worth deciding separately whether `--wallet-provider` should be **rejected** for `farcaster-miniapp` and `ai-chat` rather than silently honoured, since those templates ship their own connector wiring. That would remove the overlap at the source, and relates to #411 and #412 on flag handling.

Issue #400, opened 2026-08-11
thirdweb template: undeclared @tanstack/react-query dependency and invalid client prop on ThirdwebProvider


GigaHierz:
## What happens

```bash
npx @celo/celo-composer@2.4.13 create demo -t basic --wallet-provider thirdweb -c none -y
cd demo && pnpm install && pnpm type-check
```
```
web:type-check: src/components/wallet-provider.tsx(5,50): error TS2307: Cannot find module '@tanstack/react-query' or its corresponding type declarations.
web:type-check: src/components/wallet-provider.tsx(13,25): error TS2322: Type '{ children: ReactNode; client: ThirdwebClient; }' is not assignable to type 'IntrinsicAttributes & { connectionManager?: ... }'.
```

## Root causes

1. **Undeclared dependency.** `templates/wallets/thirdweb/components/wallet-provider.tsx.hbs:5` imports `@tanstack/react-query`, but the thirdweb branch of `templates/base/apps/web/package.json.hbs` (lines 25–26) adds only `"thirdweb": "^5.0.0"`. The rainbowkit branch declares `@tanstack/react-query` explicitly; the thirdweb branch relies on it resolving as a phantom dependency through pnpm's store, which fails type-checking and is fragile at runtime.

2. **Wrong provider API.** `wallet-provider.tsx.hbs:13` passes `client={client}` to `<ThirdwebProvider>`. In thirdweb v5 the provider takes no `client` prop — the client is passed to hooks/components like `ConnectButton`. Notably, the ai-chat template's own `components/thirdweb-client-provider.tsx` already does this correctly (`<ThirdwebProvider>` with no props), so the two copies of the same idea disagree.

(For completeness: `pnpm build` on this scaffold currently fails even earlier with `Module not found: Can't resolve '@x402/evm'` pulled through `thirdweb → @base-org/account → @coinbase/cdp-sdk` — the same dependency-drift class as #398, just through the thirdweb chain instead of wagmi/rainbowkit.)

## Fix

- Add `@tanstack/react-query` to the thirdweb branch of `templates/base/apps/web/package.json.hbs`.
- Remove the `client` prop from `<ThirdwebProvider>` in `templates/wallets/thirdweb/components/wallet-provider.tsx.hbs`.

Reproduced with @celo/celo-composer 2.4.13 (main @ 710dd89), Node 20.19, pnpm 10.

Issue #399, opened 2026-08-11
Generated Hardhat config is invalid TypeScript — unquoted celo-sepolia object keys break compile/test/deploy


GigaHierz:
## What happens

Every project scaffolded with Hardhat (the default for `basic` and `minipay`) fails on any Hardhat task, because the generated `hardhat.config.ts` is not valid TypeScript.

```bash
npx @celo/celo-composer@2.4.13 create demo -y   # basic + rainbowkit + hardhat
cd demo && pnpm install && pnpm contracts:compile
```
```
hardhat:compile: hardhat.config.ts(23,12): error TS1003: Identifier expected.
hardhat:compile: hardhat.config.ts(23,58): error TS1005: ':' expected.
hardhat:compile: hardhat.config.ts(24,24): error TS1005: ',' expected.
...
hardhat:compile: hardhat.config.ts(61,1): error TS1128: Declaration or statement expected.
Failed:    hardhat#compile
```

## Root cause

`templates/contracts/hardhat/hardhat.config.ts.hbs` uses `celo-sepolia` as a bare object key, twice:

```ts
// line 22
    celo-sepolia: {
      url: "https://forno.celo-sepolia.celo-testnet.org/",
// line 36
      celo-sepolia: process.env.ETHERSCAN_API_KEY || "",
```

A hyphenated identifier parses as `celo - sepolia` (a subtraction expression), so the file fails to parse. Line 48 inside `customChains` is already correctly quoted (`network: "celo-sepolia"`), which suggests an incomplete Alfajores → Celo Sepolia rename.

## Impact

`pnpm build`, `pnpm contracts:compile`, `contracts:test`, `contracts:deploy:celo-sepolia`, and `contracts:deploy:celo` all fail out of the box — i.e. every contracts command documented in `docs/integrations/contracts/hardhat.mdx`, `docs/development/workflow.mdx`, and `docs/development/deployment.mdx`.

## Fix

Quote both keys: `"celo-sepolia": {` and `"celo-sepolia": process.env.ETHERSCAN_API_KEY || ""`.

Reproduced with @celo/celo-composer 2.4.13 (main @ 710dd89), Node 20.19, pnpm 10.

PR #447, opened 2026-08-12
test(cli): scaffold every combination and assert on the output

### LINKED ISSUE #414: CLI has zero tests and jest --passWithNoTests gates prepublishOnly — releases ship unverified
## Summary

The CLI has zero tests, and the test script is wired so that this passes silently — including as a release gate.

## Details

- `jest.config.cjs:3-5` sets `testMatch: ['<rootDir>/src/**/*.test.ts']`; there are no `*.test.ts` files anywhere under `src/`. (The only test files in the repo belong to the vendored `templates/ai/chat-template` app and are excluded.)
- `package.json` `"test": "jest --passWithNoTests"` therefore always exits 0:
  ```
  > jest --passWithNoTests
  No tests found, exiting with code 0
  ```
- That no-op is a gate in `prepublishOnly` (`package.json:15`) and in `scripts/release.sh:47` — releases are effectively untested. Several currently-open scaffold-breaking bugs (e.g. the hardhat config syntax error, #396) would have been caught by a single "scaffold each template and type-check it" smoke test.
- The jest config is also not usable as-is: the project is `"type": "module"` with `.js`-suffixed relative imports, but `jest.config.cjs` has no ESM configuration (`useESM`, `extensionsToTreatAsEsm`, `moduleNameMapper` for `.js` specifiers), so the first test anyone writes will fail to resolve imports.
- `ts-jest`, `jest`, `@types/jest` are installed solely to support the no-op; `tsconfig.json` excludes a `tests` directory that doesn't exist.

## Suggested fix

Add an ESM-ready jest (or vitest) setup plus a smoke suite that scaffolds each template/wallet/contract combination into a temp dir and runs `tsc --noEmit` (and ideally `next build` in CI). Drop `--passWithNoTests` once real tests exist.

ghozzza:
Closes #414.

## Two problems, and the second one explains the first

**The gate was a no-op.** `jest --passWithNoTests` always exits 0, and it sits in `prepublishOnly` and `scripts/release.sh:47` — so every release shipped through a green check that verified nothing.

**And a real test could not have run anyway.** The project is `"type": "module"` with `.js` specifiers on relative imports; jest resolves those literally. Probed it before changing anything:

```
$ cat > src/utils/__probe.test.ts
  import { validateProjectName } from "./validation.js";
$ npx jest src/utils/__probe.test.ts
  Cannot find module './validation.js' from 'src/utils/__probe.test.ts'
```

A `moduleNameMapper` that strips the extension for resolution fixes it; the same probe then passes. That is worth knowing, because it means "nobody wrote tests" was partly "nobody could".

## What the suite does

It runs **the CLI**, not its parts, because that is where every bug in this review round lived — each one invisible to a unit test and obvious in a generated project.

Seven combinations — basic, basic+thirdweb, basic+none+none, basic+foundry, minipay, farcaster-miniapp, ai-chat — each asserted on three things:

1. **It scaffolds**, and every generated `.ts`/`.tsx` parses (TS1xxx only).
2. **No `package.json` declares the same dependency twice.** `JSON.parse` keeps the last, so a doubled key silently discards a version.
3. **Every `@/components/x` import names a component that exists *and exports the name used*.** Checking the file exists is not enough — in #396 the file existed and exported a different name, which is exactly how it survived review.

**Dependencies are deliberately not installed.** That would be minutes per combination, and every bug above is visible without them. The cost is that a missing import cannot be distinguished from an uninstalled one, so the TypeScript check reads syntax errors only — which is what #399 was.

## Five cases are `it.failing`, on purpose

| case | issue | fix |
|---|---|---|
| `-t ai-chat` | #387 — `templates/` resolves to `dist/templates` | #436 |
| duplicate deps | #400, #429 — react-query, viem, wagmi declared twice | #428 |
| `-t basic`, `+thirdweb`, `+foundry` component imports | #396 — navbar imports `ConnectButton`, file exports `WalletConnectButton` | #397 |

`it.failing` passes while the bug exists and **fails once it is fixed**, so the marker cannot outlive the defect — whoever merges the fix is forced to delete it.

**Verified that it actually works that way**: applying #397's diff locally turns those three red immediately.

**That this list is not empty is the argument for the suite.** Five real, currently-open defects, all of which a release would have shipped through the old green check.

## Also

`"test": "jest --passWithNoTests"` → `"test": "jest"`. `prepublishOnly` and `release.sh` now run something.

## Relationship to #393

#393 adds `template-wiring.test.ts`, which passes today only because it imports nothing from `src/` — it shells out to the CLI. Both suites coexist; this one adds the resolver fix that makes *any* src-importing test possible.

### DIFF
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -8,7 +8,18 @@ module.exports = {
     '<rootDir>/node_modules/',
     '<rootDir>/dist/'
   ],
+  // The project is "type": "module" and its TypeScript sources use `.js`
+  // specifiers on relative imports, which is what NodeNext requires. Jest
+  // resolves those literally and fails with "Cannot find module './x.js'",
+  // so any test importing a src module could not run at all — which is part
+  // of why there were none. Strip the extension for resolution only.
+  moduleNameMapper: {
+    '^(\\.{1,2}/.*)\\.js$': '$1'
+  },
   transform: {
     '^.+\\.ts$': 'ts-jest'
-  }
+  },
+  // Scaffolding a project runs the real CLI end to end: plop, template
+  // rendering, git init. The default 5s is not enough on a cold cache.
+  testTimeout: 120000
 };
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build": "tsc",
     "dev": "tsx src/index.ts",
     "start": "tsc && node dist/index.js",
-    "test": "jest --passWithNoTests",
+    "test": "jest",
     "lint": "eslint src --ext .ts",
     "clean": "rimraf dist",
     "prepublishOnly": "pnpm clean && pnpm build && pnpm test",
--- a/src/generators/scaffold-smoke.test.ts
+++ b/src/generators/scaffold-smoke.test.ts
@@ -1,0 +1,180 @@
+/**
+ * Smoke tests: scaffold real projects and assert on what comes out.
+ *
+ * The CLI's failure mode is not a crash — it is generating a project that looks
+ * fine and does not work. Every scaffold-breaking bug found in the review this
+ * suite was written for was invisible to a unit test and obvious to a generated
+ * project:
+ *
+ *   - hardhat.config.ts was not valid TypeScript, so every hardhat scaffold
+ *     failed on any task (#399)
+ *   - the navbar imported a component nothing wrote (#396, #403)
+ *   - package.json shipped a duplicate dependency key (#400)
+ *   - ai-chat could not be scaffolded at all (#387)
+ *
+ * So these tests run the CLI, not its parts. They deliberately do NOT install
+ * dependencies: that takes minutes per combination, and the bugs above are all
+ * visible without it. What that costs is type resolution — a missing import
+ * cannot be distinguished from an uninstalled one — so the TypeScript check
+ * below looks only at SYNTAX errors (TS1xxx), which need no node_modules.
+ */
+
+import { execFileSync } from "child_process";
+import { mkdtempSync, rmSync, existsSync, readFileSync, readdirSync, statSync } from "fs";
+import { tmpdir } from "os";
+import { join, resolve } from "path";
+
+// __dirname, not import.meta: ts-jest compiles these to CJS, matching how
+// template-wiring.test.ts resolves the same path.
+const REPO_ROOT = resolve(__dirname, "..", "..");
+const CLI = join(REPO_ROOT, "dist", "index.js");
+
+let workdir: string;
+
+beforeAll(() => {
+  // dist/ is what users run, so that is what gets tested.
+  execFileSync("npm", ["run", "build"], { cwd: REPO_ROOT, stdio: "pipe" });
+  workdir = mkdtempSync(join(tmpdir(), "celo-composer-smoke-"));
+});
+
+afterAll(() => {
+  if (workdir) rmSync(workdir, { recursive: true, force: true });
+});
+
+function scaffold(name: string, args: string[]): string {
+  execFileSync("node", [CLI, "create", name, ...args, "--skip-install", "-y"], {
+    cwd: workdir,
+    stdio: "pipe",
+  });
+  return join(workdir, name);
+}
+
+/** Every .ts/.tsx file under a directory, ignoring anything vendored. */
+function typescriptFiles(root: string): string[] {
+  const out: string[] = [];
+  const walk = (dir: string): void => {
+    for (const entry of readdirSync(dir)) {
+      if (entry === "node_modules" || entry === ".git" || entry === ".next") continue;
+      const full = join(dir, entry);
+      if (statSync(full).isDirectory()) walk(full);
+      else if (/\.tsx?$/.test(entry)) out.push(full);
+    }
+  };
+  walk(root);
+  return out;
+}
+
+/**
+ * Syntax errors only. Without node_modules every import is unresolved, so
+ * TS2307 and friends say nothing — but TS1xxx means the file does not parse,
+ * which is a template bug every time.
+ */
+function syntaxErrors(projectPath: string): string[] {
+  const files = typescriptFiles(projectPath);
+  if (files.length === 0) return [];
+  try {
+    execFileSync("npx", ["tsc", "--noEmit", "--skipLibCheck", "--target", "es2022",
+      "--module", "esnext", "--moduleResolution", "bundler", "--jsx", "preserve", ...files],
+      { cwd: projectPath, stdio: "pipe" });
+    return [];
+  } catch (err) {
+    const output = String((err as { stdout?: Buffer }).stdout ?? "");
+    return output.split("\n").filter((line) => /error TS1\d{3}:/.test(line));
+  }
+}
+
+/**
+ * Combinations this suite currently catches as broken on `main`, each with the
+ * PR that fixes it. Written as `it.failing`, which passes while the bug exists
+ * and FAILS once it is fixed — so the marker removes itself rather than
+ * quietly outliving the bug.
+ *
+ * That this list is not empty is the point: every entry is a real defect this
+ * suite would have caught before release.
+ */
+const KNOWN_BROKEN: Record<string, string> = {
+  "-t ai-chat": "#387 — templates/ resolves to dist/templates, so ai-chat cannot scaffold (fix: #436)",
+  "duplicate-deps": "#400 / #429 — react-query, viem and wagmi declared twice (fix: #428, plus #429 filed)",
+  "-t basic": "#396 — navbar imports ConnectButton; the file exports WalletConnectButton (fix: #397)",
+  "-t basic --wallet-provider thirdweb": "#396 — same (fix: #397)",
+  "-t basic -c foundry": "#396 — same (fix: #397)",
+};
+
+const COMBINATIONS = [
+  { name: "smoke-basic", args: ["-t", "basic"] },
+  { name: "smoke-basic-thirdweb", args: ["-t", "basic", "--wallet-provider", "thirdweb"] },
+  { name: "smoke-basic-none", args: ["-t", "basic", "--wallet-provider", "none", "-c", "none"] },
+  { name: "smoke-foundry", args: ["-t", "basic", "-c", "foundry"] },
+  { name: "smoke-minipay", args: ["-t", "minipay"] },
+  { name: "smoke-farcaster", args: ["-t", "farcaster-miniapp"] },
+  { name: "smoke-aichat", args: ["-t", "ai-chat"] },
+];
+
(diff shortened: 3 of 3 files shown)


GigaHierz (changes requested):
The harness design (scaffold matrix, self-removing `it.failing` markers, ESM jest fix, dropping `--passWithNoTests`) is worth landing. But the suite's central assertion is currently a no-op, and this was verified empirically on the current head:

**The TypeScript check never checks TypeScript.** `syntaxErrors()` runs `execFileSync("npx", ["tsc", ...])` with `cwd` set to the scaffolded project — which has no `node_modules` — so npx resolves to the npm registry's placeholder package literally named `tsc`, which prints "This is not the tsc command you are looking for" and exits 1 with no `error TS…` lines. The catch block greps stdout, finds nothing, returns `[]`. Three demonstrations against this branch:

- The full suite passes 15/15 — including plain-`it` `smoke-basic` — over a scaffold whose `apps/contracts/hardhat.config.ts` contains **21 real TS1xxx syntax errors** (the unquoted `celo-sepolia:` keys, issue #399, still on main). The header comment's claim that the suite would have caught #399 is not true of the code as written.
- Running the test's exact npx command from a scaffold reproduces the placeholder banner.
- Offline (registry unreachable), npx fails on **stderr**, the catch reads only `err.stdout`, and the check still silently passes.

Fix: use the repo's own `typescript` package programmatically — `ts.createSourceFile` per generated `.ts`/`.tsx` and collect parse diagnostics. No npx, no network, faster, and it turns `smoke-basic` red on #399 exactly as intended (add a `KNOWN_BROKEN` entry for it until #427 merges).

Secondary items, all still present:

- The component check validates import↔export pairs but not used-but-never-imported JSX: minipay's generated navbar renders `<WalletConnectButton />` with no import (the #396 bug) and passes green.
- `execFileSync("npm"/"npx", …)` without `shell: true` fails on Windows (`.cmd` shims) — and with no CI workflows, contributor machines are where this runs.
- The duplicate-key regex `/^\s{4}"…":/gm` is section-blind: on a hardhat scaffold, `typechain` (script name + devDependency) false-positives. Today's hardcoded manifests dodge it; any coverage extension trips it.
- The third describe silently passes vacuously if the first describe's scaffold is missing (`existsSync` short-circuit) — should fail loudly instead.
- `KNOWN_BROKEN["duplicate-deps"]` is dead data (never matches a label), and `KNOWN_BROKEN[label] && label === "-t ai-chat"` is just `label === "-t ai-chat"`.

Given `pnpm test` now gates `prepublishOnly`, an inert headline check is worse than none — it certifies broken scaffolds as release-ready. Happy to re-verify quickly once the parse check is rewritten.

ghozzza:
Rewritten at `8c34b23`. You were right, and it was inert for **two** reasons, not one.

## The headline check

Reproducing it here failed differently from your run, which makes the point stronger:

```
$ npx tsc --noEmit --skipLibCheck … <files>    # cwd = scaffold, no node_modules
error TS5112: tsconfig.json is present but will not be loaded if files are
              specified on commandline. Use '--ignoreConfig' to skip this error.
```

On your machine npx fetched the registry placeholder; on mine a `tsc` was reachable and refused to run because the scaffold ships a `tsconfig.json`. **Both produce zero `error TS…` lines on stdout, so the catch filtered an empty string and returned `[]`.** The check's result depended on what happened to be installed on the runner, and on neither machine did it check anything.

Now parsed in-process with **`ts.transpileModule`** — public API, single file, no module resolution, no config file, no network. It reports the same parse diagnostics: **21 on `hardhat.config.ts`**, matching the count in #399, and 0 on a healthy file.

**Scoped from measurement, not assumption.** Generating all seven combinations shows #399 hits exactly three of them:

| combination | parse diagnostics |
|---|---|
| `-t basic` | **21** — `apps/contracts/hardhat.config.ts` |
| `-t basic --wallet-provider thirdweb` | **21** — same |
| `-t minipay` | **21** — same |
| `-t basic --wallet-provider none -c none` | 0 |
| `-t basic -c foundry` | 0 |
| `-t farcaster-miniapp` | 0 |

Those three now carry `KNOWN_BROKEN` entries naming #399 and #427, as you suggested.

**And the markers self-clean.** Merging #427 into this branch and re-running:

```
✕ -t basic                              Failing test passed even though it was supposed to fail.
✕ -t basic --wallet-provider thirdweb   Remove `.failing` to remove error.
✕ -t minipay
✓ -t basic --wallet-provider none -c none
✓ -t basic -c foundry
✓ -t farcaster-miniapp
```

Three red, three untouched — so the entries are scoped to the defect rather than to the suite.

## The five secondary items

**Used-but-never-imported JSX.** Walking the AST instead of only matching import statements. On a minipay scaffold it reports exactly what you named, and nothing else:

```
apps/web/src/components/navbar.tsx -> WalletConnectButton
```

Farcaster comes back clean, so it is not just flagging everything. `-t minipay` joins the components `KNOWN_BROKEN` list for #396.

**Duplicate-key regex.** Replaced with a scanner that tracks brace depth and counts keys per object. Your `typechain` example, on a real hardhat scaffold — script at line 15, devDependency at line 45:

```
apps/contracts/package.json  new scanner -> []
                             old regex   -> ["typechain"]
```

The reason it never fired is the same reason it needed fixing: the test only looked at the root and `apps/web` manifests. Manifests are now found by walking the scaffold, so `apps/contracts` is covered — and that is exactly the coverage extension that would have tripped the old check.

**Windows.** `npm.cmd` on `win32`. The `npx` call is gone entirely, so that was the only shim left.

**Vacuous third describe.** The `existsSync` short-circuit is gone. `apps/web/src` missing is now a loud failure, except for `-t ai-chat` where its *absence* is asserted, since a flat app must not have one.

**Dead `KNOWN_BROKEN` data.** Restructured to `{ suite, label, reason }` and consulted through one `runnerFor(suite, label)`. `"duplicate-deps"` matched no label and is now a real entry against the manifests suite; the `KNOWN_BROKEN[label] && label === "-t ai-chat"` tautology is gone.

15/15 on this branch, and the three that should go red do go red once #427 lands. Happy to have it re-checked.

ghozzza:
`#427` merging did the thing this suite was built to do, so an update at `7825412`.

Merged current `main` into the branch and the three `#399` entries went red on the next run, unprompted:

```
✕ -t basic                              Failing test passed even though it was
✕ -t basic --wallet-provider thirdweb   supposed to fail. Remove `.failing`.
✕ -t minipay
```

Exactly the three that ship a hardhat package, and nothing else moved. They are removed, with a note in their place recording that the list is meant to shrink this way rather than be pruned by hand. **15/15 against current `main`.**

Worth saying plainly since it is the argument for the marker style: nobody had to remember this file existed. The fix landed, the suite objected, and the objection named the exact combinations and the reason. Had the entries been plain comments they would still be sitting there claiming a bug that no longer exists.

One incidental: `-t basic -c foundry` went from ~300 ms to ~2.2 s now that #430 makes Foundry projects real. Still well inside the budget, flagging only so a future slowdown is not mistaken for this.

GigaHierz (approved):
Re-verified at 8c34b23 by running the suite in a clean clone: 15/15 green, and the decisive check now works — unmarking the basic scaffold produces the 21 real TS1xxx diagnostics from the hardhat config (issue #399), and the new JSX walker reports exactly ['WalletConnectButton'] on the minipay navbar (#396). Duplicate-key scanner, vacuous-pass fix, and KNOWN_BROKEN restructure all verified. One remaining nit to fix on merge: execFileSync('npm.cmd', ...) without shell:true throws EINVAL on Windows under CVE-2024-27980-patched Node — add shell:true (or cmd /c). Note the KNOWN_BROKEN markers will need updating as the fix PRs land (#427/#430/#436 already merged).

GigaHierz:
Now that #427, #430, #431, #435, #436, #437, #397, #394, #428, #434, #449 have merged, most KNOWN_BROKEN markers reference fixed bugs — jest will fail loudly on each it.failing that now passes. Please rebase onto current main, prune the markers for merged fixes (keep #400-adjacent ones only if the duplicate still renders — #450 is still pending), and add shell:true to the npm.cmd execFileSync for Windows. Then this merges as the final gate.

ghozzza:
Rebased at `f73b566`. All three, plus one thing the rebase exposed.

## The markers pruned themselves, and named which

Merging main turned seven of the eight red. Removed: `-t ai-chat` (#436), and the four component entries `-t basic`, `-t basic --wallet-provider thirdweb`, `-t basic -c foundry`, `-t minipay` (#397), on top of the three #399 entries already dropped when #427 landed.

**One stays**, exactly as you called it: `-t farcaster-miniapp --wallet-provider rainbowkit` still renders the duplicate, because #428 fixed only the react-query half and #450 is still open for `viem`/`wagmi`. Its reason line now says that rather than pointing at both PRs as if both had landed. **15/15.**

## `shell: true` — done, and scoped

`execFileSync(NPM, …, { shell: process.platform === "win32" })`. Conditional rather than unconditional: passing `shell: true` on POSIX re-parses the argv through `/bin/sh`, which is a quoting hazard for no gain when `npm` resolves fine there. The `npm.cmd` name stays for the same platform.

## The rebase found a real bug in my own check

`-t ai-chat` did not go red like the others. It kept failing, and not for the reason the marker claimed — #436 had fixed the scaffolding. The check itself was crashing:

```
Error: Debug Failure. Output generation failed
    at transpileWorker (typescript.js:144512)
```

`transpileModule` parses **and emits**. A declaration file produces no output, so emit throws. Every ai-chat scaffold ships `next-env.d.ts`, so the one template that check was written to protect was the one it could not run on. One file out of 148.

Swapped to `ts.createSourceFile` and its `parseDiagnostics`, which parses and stops:

```
transpileModule    1 crash, 0 files reported
createSourceFile   0 crashes, 148 files parsed, 0 diagnostics
```

ai-chat is genuinely clean; the crash was hiding that rather than reporting it. `parseDiagnostics` needs a cast since it is not on the public `SourceFile` type, which is noted at the call site.

Worth saying since it undercuts my own earlier argument: I chose `transpileModule` in the last round *because* it was the public API. It was the public API and the wrong tool.

GigaHierz:
Rebased onto main as a maintainer edit and retired the final KNOWN_BROKEN marker (#429's viem/wagmi duplicate — #450 just merged). The Windows shell fix was already on the branch. Suite runs 15/15 green with an empty markers list: every combination scaffolds, every generated file parses with the real TypeScript API, every manifest is duplicate-free. With this merged, prepublishOnly finally gates releases on something real.

GigaHierz (approved):
Approving post-rebase: the in-process TypeScript check was previously verified to catch real syntax errors (21 diagnostics on the pre-#427 hardhat config), the JSX walker catches the #396 shape, and the suite now runs 15/15 with zero expected-failure markers.

Issue #415, opened 2026-08-11
Cleanup: dead Plop actions/prompts, unused deps (path polyfill, plop, cross-env), non-portable commandExists, stale release.sh


GigaHierz:
## Summary

Grouped cleanup issue: dead code, unused dependencies, and drifted tooling in the CLI itself. None of these break users on their own, but together they add noise, install weight, and misleading signals.

## Dead code

- `src/plopfile.ts:59-68` — the "copy static assets from base" action globs `templates/base/**/*` excluding `.hbs`; all 21 files in `templates/base` are `.hbs`, so the action matches nothing. (The equivalent farcaster action at lines 238-256 does work.)
- `src/plopfile.ts:24-45` — two generator `prompts` (`projectName`, `description`) that can never fire: the code always calls `generator.runActions(...)` directly, never `runPrompts`.
- `src/generators/plop-generator.ts:108-111` — `runPlopGenerator` is exported but never called.
- `PlopConfig.installDependencies` (`plop-generator.ts:16`) and `PlopData.installDependencies` (`plopfile.ts:11`) are declared but never read.
- `CreateOptions.templateType` and `CreateOptions.miniapp*` (`src/commands/create.ts:14,19-22`) have no corresponding commander flags — they're read in several places but can never be set from the CLI.
- `src/generators/project-generator.ts:70,86,140` — unused `catch (error)` bindings.

## Dependencies

- `path` `^0.12.7` (dependencies) — this is the deprecated browserify polyfill; Node always resolves bare `"path"` to the core module, so the package is dead weight with transitive shims.
- `plop` `^4.0.1` (dependencies) — imported type-only (`import { NodePlopAPI } from "plop"`); TS elides it, so it belongs in devDependencies. `node-plop` is the actual runtime engine.
- `cross-env` (devDependencies) — zero references anywhere.

## Tooling drift

- `scripts/release.sh:63-70` promises "The GitHub Action will now… Publish to npm" and links to `/actions`, but the repo has no `.github/` directory / workflows. It also uses `npm version` in a pnpm-managed repo.
- `commandExists()` (`src/commands/create.ts:406-412`) spawns `command -v` without `shell: true`. `command` is a shell builtin; this happens to work on macOS (which ships a `/usr/bin/command` binary) but fails on systems without one, making the "Foundry already installed" check non-portable. The sibling `foundryup` spawn (`create.ts:438-447`) registers no `error` handler, so a missing binary raises an unhandled `error` event.
[...]

PR #448, opened 2026-08-12
chore(cli): drop dead dependencies, dead code and stale tooling claims

ghozzza:
Refs #415 — the parts that do not collide with the open queue. Scope note at the bottom.

## Dependencies

| package | why it goes |
|---|---|
| `path` `^0.12.7` | the deprecated **browserify polyfill**. Node resolves bare `"path"` to the core module, so this and its transitive shims were installed for nothing |
| `plop` `^4.0.1` | imported **type-only**. Confirmed from the build: the only reference in `dist/` is `plopfile.d.ts`. Moved to devDependencies; `node-plop` is the runtime engine and stays |
| `cross-env` | zero references anywhere in the repo |

Also changed the plop import to `import type`, so the elision is guaranteed by the syntax rather than left to TypeScript noticing.

## Dead code

- **`runPlopGenerator`** — exported, never called.
- **Three `catch (error)` bindings** that nothing reads. Each block was checked for a reference to `error` first, so any that use it were left alone.

## Tooling

**`release.sh` promised something that cannot happen.** It ended with:

> 📦 The GitHub Action will now: Run CI tests · Build the project · **Publish to npm** · Create GitHub release
> 🔗 Monitor the progress at: …/actions

There is no `.github/` directory. Nothing runs, nothing publishes, and the script's last words tell you to go watch it happen. A release silently stopped at the tag. It now says so and prints the actual publish command.

Also `npm version` → `pnpm version`, in a repo whose `packageManager` is pnpm; and `tsconfig.json` no longer excludes a `tests/` directory that does not exist.

`.gitignore` was two lines — added `*.tsbuildinfo`, `.DS_Store` and `.prune-stash/`.

## Verified

Every template still scaffolds with those dependencies removed:

```
basic   ok      minipay   ok      farcaster-miniapp  ok      basic+foundry  ok
ai-chat FAILED  ← ENOENT on dist/templates/ai/chat-template
```

That last one is **#387**, present on `main` and fixed in #436 — not this change. Confirmed by reading the error path rather than assuming.

## Deliberately out of scope

The issue also lists dead Plop actions and prompts in `src/plopfile.ts`, and `CreateOptions` fields with no commander flags in `src/commands/create.ts`.

**Four open PRs already touch those two files** — #393, #394 on the plopfile; #393, #430, #440 on create.ts. Deleting lines in the middle of that queue buys a small cleanup for a set of avoidable conflicts. Happy to take them as a follow-up once those land, and the issue can stay open for it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,10 @@
 node_modules
 .env
 dist
+
+# build artefacts and editor noise
+*.tsbuildinfo
+.DS_Store
+
 # leftover from an interrupted pnpm pack (older versions stashed here)
 .prune-stash/
--- a/package.json
+++ b/package.json
@@ -50,9 +50,7 @@
     "fs-extra": "^11.1.1",
     "inquirer": "^9.2.12",
     "node-plop": "^0.32.0",
-    "ora": "^7.0.1",
-    "path": "^0.12.7",
-    "plop": "^4.0.1"
+    "ora": "^7.0.1"
   },
   "devDependencies": {
     "@types/fs-extra": "^11.0.4",
@@ -61,9 +59,9 @@
     "@types/node": "^20.8.0",
     "@typescript-eslint/eslint-plugin": "^6.7.0",
     "@typescript-eslint/parser": "^6.7.0",
-    "cross-env": "^7.0.3",
     "eslint": "^8.50.0",
     "jest": "^29.7.0",
+    "plop": "^4.0.1",
     "rimraf": "^5.0.5",
     "ts-jest": "^29.4.1",
     "tsx": "^3.14.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -26,12 +26,6 @@ importers:
       ora:
         specifier: ^7.0.1
         version: 7.0.1
-      path:
-        specifier: ^0.12.7
-        version: 0.12.7
-      plop:
-        specifier: ^4.0.1
-        version: 4.0.1
     devDependencies:
       '@types/fs-extra':
         specifier: ^11.0.4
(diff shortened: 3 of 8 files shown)


GigaHierz:
Conflict with #438's .gitignore change as anticipated — 30-second rebase (both add .prune-stash/; keep one entry plus your *.tsbuildinfo/.DS_Store additions). Approved once rebased. I've also updated the PR body's 'Closes #415' to 'Refs #415' so the issue stays open for the deferred plopfile/CreateOptions items.

ghozzza:
Rebased at `3c9d915`. `.gitignore` now carries one `.prune-stash/` entry plus the `*.tsbuildinfo` and `.DS_Store` additions, as you described. `tsc` clean.

Noted on `Refs #415` rather than `Closes` — that is the right call, the plopfile dead actions and the flagless `CreateOptions` fields are still outstanding and were deliberately left because other PRs were touching those two files. Those are now merged, so that remainder is unblocked whenever you want it picked up.

GigaHierz:
Rebased onto main as a maintainer edit — resolved the .gitignore conflict with the union of both changes (#438's interrupted-pack comment + this PR's *.tsbuildinfo/.DS_Store additions, single .prune-stash/ entry). Re-verified: install with regenerated lockfile, build, lint all pass, and a minipay scaffold generates through the node-plop path (confirming the plop-to-devDependencies move is safe at runtime). Body already updated to 'Refs #415'.

GigaHierz (approved):
Approving post-rebase — verification unchanged from the original review (every removal confirmed safe by building and scaffolding), and the rebase only resolved the known .gitignore overlap.

Issue #402, opened 2026-08-11
farcaster-miniapp: next build fails — shipped .eslintrc.json references @typescript-eslint rules without installing the plugin


GigaHierz:
## What happens

Every `farcaster-miniapp` scaffold fails `pnpm build`:

```bash
npx @celo/celo-composer@2.4.13 create demo -t farcaster-miniapp -y
cd demo && pnpm install && pnpm build
```
```
web:build: ./src/lib/env.ts
web:build: 1:1  Error: Definition for rule '@typescript-eslint/no-unused-vars' was not found.  @typescript-eslint/no-unused-vars
web:build: ./src/lib/memory-store.ts
web:build: 1:1  Error: Definition for rule '@typescript-eslint/no-unused-vars' was not found.  @typescript-eslint/no-unused-vars
... (every file)
web:build:  ELIFECYCLE  Command failed with exit code 1.
Failed:    web#build
```

`pnpm type-check` passes — the failure is purely the ESLint step that `next build` runs.

## Root cause

`templates/farcaster-miniapp/apps/web/.eslintrc.json.hbs` configures rules from `@typescript-eslint` (e.g. `@typescript-eslint/no-unused-vars`), but neither `@typescript-eslint/eslint-plugin` nor a config that provides it is added to the farcaster branch of `templates/base/apps/web/package.json.hbs`. `eslint-config-next` alone does not register those rule definitions for this usage.

Related context: #395 covers the *other* templates failing lint because they ship no ESLint config at all; farcaster-miniapp is the one template that ships a config, and its config is broken in a different way.

## Fix

Either add the `@typescript-eslint` plugin/parser to the farcaster dependency branch, or drop the `@typescript-eslint/*` rule overrides from the shipped `.eslintrc.json.hbs`.

Reproduced with @celo/celo-composer 2.4.13 (main @ 710dd89), Node 20.19, pnpm 10.

PR #449, opened 2026-08-12
fix(templates): migrate the farcaster mini app off the deprecated frame SDK

### LINKED ISSUE #405: farcaster-miniapp: half-migrated SDK layer — deprecated frame-sdk/frame-core, "latest" pins, mixed addFrame/addMiniApp, legacy v1 webhook
## Summary

The farcaster-miniapp template's SDK layer is a half-finished migration from Frames to Mini Apps: deprecated package names, unpinned `"latest"` dependencies, inconsistent SDK calls, and a webhook handler for a protocol version that no longer fires.

## Details

1. **Deprecated packages.** `templates/base/apps/web/package.json.hbs:27-30` (farcaster branch) depends on `@farcaster/frame-sdk` (`^0.0.61`) and `@farcaster/frame-core` — both deprecated in favor of `@farcaster/miniapp-sdk` / `@farcaster/miniapp-core` — while simultaneously using the new `@farcaster/miniapp-wagmi-connector`.

2. **`"latest"` pins.** `@farcaster/frame-core`, `@farcaster/miniapp-wagmi-connector`, and `@farcaster/quick-auth` are pinned to `"latest"`, so every scaffold resolves different versions and builds are unreproducible.

3. **Inconsistent SDK usage.** `contexts/miniapp-context.tsx.hbs:58` calls the deprecated `sdk.actions.addFrame()` while `templates/base/apps/web/src/app/page.tsx.hbs:129` calls `sdk.actions.addMiniApp()` — same SDK, two generations of API in one project. The context also stubs types away (`type FrameContext = any; type AddFrameResult = any;`).

4. **Types imported from the wrong package.** `lib/notification-client.ts.hbs` and `lib/memory-store.ts.hbs` import `FrameNotificationDetails` / `SendNotificationRequest` / `sendNotificationResponseSchema` from `@farcaster/frame-sdk`; these live in the `-core` package, which is declared as a dependency but never imported.

5. **Legacy webhook.** `apps/web/src/app/api/webhook/route.ts.hbs` parses the Frames-v1 shape (`untrustedData.buttonIndex` / `trustedData.messageBytes`). Mini App webhook events (`frame_added`, `notifications_enabled`, …) are JFS-signed JSON — the shipped handler will never process a real event, so the notification flow (`lib/notification-client.ts.hbs`, `api/notify`) is dead on arrival.

## Suggested fix

Migrate to `@farcaster/miniapp-sdk` / `@farcaster/miniapp-core` with exact-ish semver pins, standardize on `addMiniApp()`, and rewrite the webhook route for the Mini App event format (e.g. using `verifyJsonFarcasterSignature` helpers from miniapp-node/core).

Found in @celo/celo-composer 2.4.13 (main @ 710dd89). Note: type-check currently passes (frame-sdk 0.0.61 ships both `addFrame` and `addMiniApp`), so this is drift/correctness debt rather than a compile failure — but the webhook path is functionally broken.

ghozzza:
Closes #405.

Every point in the issue, plus two things the migration exposed that the issue could not have known about.

## The five points

**1. Deprecated packages.** `@farcaster/frame-sdk` and `@farcaster/frame-core` are gone. In their place:

| | version | why it is here |
|---|---|---|
| `@farcaster/miniapp-sdk` | `^0.3.0` | client — `sdk.context`, `sdk.actions.*` |
| `@farcaster/miniapp-core` | `^0.6.0` | shared types and schemas |
| `@farcaster/miniapp-node` | `^0.2.0` | webhook verification (new) |

**2. The `"latest"` pins.** All four are now real ranges — `miniapp-wagmi-connector` at `^2.0.0` and `quick-auth` at `^0.0.8`, which are the versions a scaffold resolved to anyway. `grep -rn '": "latest"' templates/` returns nothing.

**3. `addFrame()` → `addMiniApp()`.** One generation of the API in one project now. `sdk.actions.addFrame` still exists in 0.3.0 as an alias, so this was not a compile error — which is exactly why it survived.

**4. Types from the right package.** `MiniAppNotificationDetails`, `SendNotificationRequest` and `sendNotificationResponseSchema` now come from `@farcaster/miniapp-core`, where they actually live. `FrameNotificationDetails` no longer exists under that name.

**5. The webhook.** Rewritten. This is the substantive part.

## The webhook

The old handler read `untrustedData.buttonIndex` and `trustedData.messageBytes` — the Frames v1 button-post shape. Mini App events do not look like that, so no real event ever reached the handler and the whole notification path was dead on arrival.

It now parses the JSON Farcaster Signature envelope with `parseWebhookEvent` and handles the four events `serverEventSchema` actually defines:

```
miniapp_added        notificationDetails optional — the user may add without enabling
miniapp_removed
notifications_enabled    notificationDetails required
notifications_disabled
```

**One correction to the issue:** it names the events `frame_added` / `notifications_enabled`. In `miniapp-core@0.6.0` the add and remove events are `miniapp_added` and `miniapp_removed`. Read from the installed `dist/schemas/events.d.ts` rather than from the docs.

**On the verifier, and the trade-off I picked.** Checking a JFS signature means proving the signing key belongs to the sender's fid, which requires a Farcaster hub. `@farcaster/miniapp-node` exports two ways to do that:

- `verifyAppKeyWithNeynar` — convenient, but it **throws** unless `NEYNAR_API_KEY` is set. In a scaffold that means every generated project has a webhook that crashes rather than one that explains itself.
- `createVerifyAppKeyWithHub(hubUrl, requestOptions)` — the primitive the Neynar helper is built on.

I used the primitive, with the hub URL and API key both read from the environment and a documented default. A project with neither configured answers `500` with `Could not verify the event` in the log, rather than throwing an opaque error from inside a dependency. Both variables are in `.env.template` with the consequence spelled out.

I did consider keeping the old on-chain path — the previous handler read `keyDataOf` from the Optimism Key Registry with plain viem, no API key. It cannot satisfy the interface: `VerifyAppKeyResult` requires `appFid` when valid, and `keyDataOf` returns only `(state, keyType)`. The app fid lives in the signer's metadata, which is emitted in an event rather than readable from that call.

## Two things the migration exposed

**`type FrameContext = any` was hiding a broken read.** With the real `Context.MiniAppContext` in place, `tsc` fails:

```
src/app/page.tsx(33,42): error TS2339: Property 'custody' does not exist on type 'UserContext'.
src/app/page.tsx(33,59): error TS2339: Property 'verifications' does not exist on type 'UserContext'.
```

`UserContext` is `fid`, `username`, `displayName`, `pfpUrl`, `location`. No addresses. So this line:

```ts
const walletAddress = address || user?.custody || user?.verifications?.[0] || null;
```

had two fallbacks that could never fire — when `address` was undefined it always reached `null`. Dropped, with a comment naming what the context does carry. The address comes from the connector the effect above it sets up.

**A build warning goes away.** The old handler imported `optimism` from `viem/chains`, which pulls the whole barrel including `tempo`:

```
⚠ Compiled with warnings
../../node_modules/.pnpm/ox@0.14.33.../ox/_esm/tempo/internal/virtualMasterPool.js
Critical dependency: the request of a dependency is an expression
```

The hub verifier needs no chain import, so a generated project now reports `✓ Compiled successfully`.

## Verification

Generated a real project each time rather than reading the diff.

| | `main` (control, `710dd89`) | this branch |
|---|---|---|
| `pnpm type-check` | passes — `any` hides the two errors | passes, with real types |
| `pnpm build` — compile | `⚠ Compiled with warnings` | `✓ Compiled successfully` |
| `pnpm build` — lint | fails, 22 × `@typescript-eslint/no-unused-vars` | **identical, same 22** |

Resolved versions in the generated project: `miniapp-sdk 0.3.0`, `miniapp-core 0.6.0`, `miniapp-node 0.2.0`, `miniapp-wagmi-connector 2.0.0`, `quick-auth 0.0.8`.

**The lint failure is #402 and is not touched here.** The control run is the point: it is the same 22 errors before and after, from `.eslintrc.json.hbs` shipping `@typescript-eslint` rules without the plugin. Fixing it here would mix two changes.

**What is not verified:** no real webhook event has been delivered end to end. That needs a published manifest on a public domain and a user adding the app, and I am not going to claim it from a passing type-check. What is verified is the shape — the events handled are the ones `serverEventSchema` defines, read from the package rather than from documentation.

## Merge order

Both overlapping PRs were test-merged into this branch; **neither conflicts**.

- **#435** also edits `miniapp-context.tsx.hbs`, but only the `FrameWalletProvider` mount. Different hunks.
- **#428** also edits `package.json.hbs`, in the react-query line below the farcaster block. Different hunks.

## Deliberately left alone

`frame-wallet-context.tsx` and `FrameWalletProvider` keep their names. Renaming them is a file move that would collide with #435 and buys nothing at runtime — worth a follow-up once that lands, not worth bundling here.


GigaHierz (approved):
Verified by installing the actual packages and compiling the rendered templates with tsc --strict against them (exit 0): all claimed exports exist at the pinned versions, parseWebhookEvent verifies the JFS signature before decoding, the four serverEventSchema event names match (the issue's frame_added naming was wrong — good catch), error-name strings match the dist typings character-for-character, and zero frame-sdk/frame-core/addFrame references remain anywhere in templates/ or docs/. The dropped custody/verifications fallbacks were dead code under the real types. Two non-blocking nits: NEYNAR_API_KEY/FARCASTER_HUB_URL bypass the t3-oss env.ts schema every other server var uses, and the welcome-notification result is discarded (a state:'error' return is silently ignored; only a thrown fetch error reaches the 500/retry path).

Issue #398, opened 2026-08-10
Default scaffold does not build — dependency drift pulls an unmet @x402/evm peer through wagmi


ghozzza:
### What happens

`next build` fails on a default scaffold. No template code is involved.

```
$ npx @celo/celo-composer create demo -t basic
$ cd demo && pnpm install && pnpm --filter web build

Module not found: Can't resolve '@x402/evm/upto/client'
Module not found: Can't resolve '@x402/evm/exact/client'

Import trace:
  @coinbase/cdp-sdk@1.55.0/_esm/actions/x402/signX402Payment.js
  @base-org/account@2.4.0/dist/interface/payment/charge.js
  @wagmi/connectors@6.2.0/dist/esm/exports/index.js
  wagmi@2.19.5/dist/esm/exports/connectors.js
  @rainbow-me/rainbowkit@2.2.11/dist/index.js
  ./src/components/wallet-provider.tsx

Build failed because of webpack errors
```

Checked: `@x402/evm@2.21.0` does export both subpaths. The package is simply **not in the generated project's tree at all** — it is an unmet peer of `@coinbase/cdp-sdk`, which pnpm's non-hoisted layout does not resolve.

### Why this is a template problem, not an upstream one

The base template pins wide ranges and ships no lockfile, so `wagmi: "^2.0.0"` resolves to 2.19.5 today and pulls a dependency chain that did not exist when the template was written. Nothing in the repo changed; the ecosystem moved.

The same drift shows up across the board:

| | base template | current |
| --- | --- | --- |
| next | `^14.0.0` | 16.3.0 |
| react | `^18.2.0` | 19.2.8 |
| turbo | `^1.10.16` | 2.10.9 |
| eslint | `^8.50.0` | 10.8.1 |
| packageManager | `pnpm@8.10.0` | pnpm 8 |

This is already shaping new work: `@x402/next` requires `next >= 16.2.6`, so a template that wants a payment middleware in `apps/web` cannot use it against the Next 14 pin.

### Options, all maintainer calls

1. **Narrow the ranges** so `pnpm install` cannot drift into an incompatible tree, and revisit deliberately.
2. **Bump the stack** — Next 16 and React 19 — which touches every template and needs its own testing pass.
3. **Ship a lockfile** with the base template so generated projects are reproducible.

Raising this rather than opening a PR, since any of the three changes every template and the choice is not mine to make. Happy to take whichever direction you pick.

PR #454, opened 2026-08-13
fix(templates): make the ai-chat template build

ghozzza:
Refs #387 (the build half — the scaffolding half was #436).

The `experimental.ppr` blocker you found is real, and it was the first of four. Each one only became visible once the one before it was cleared, so the fix is a chain rather than a line.

## 1. `experimental.ppr` needs canary; the pin is stable

```
Error: The experimental feature "experimental.ppr" can only be enabled
       when using the latest canary version of Next.js.
```

Removed from `next.config.ts`. Also removed `export const experimental_ppr = true` from `app/(chat)/layout.tsx`, which is the route-level opt-in for the same feature — it does not error with PPR off, Next simply ignores it, so it would have sat there as dead config pointing at a flag the template no longer sets.

Removing rather than moving to canary, as you suggested: a template that only builds on a canary release is not a starter kit.

## 2. The `@x402` optional peers reach this template too

With PPR cleared, the build failed on five unresolved modules:

```
Can't resolve '@x402/core/client'
Can't resolve '@x402/evm'
Can't resolve '@x402/evm/exact/client'
Can't resolve '@x402/evm/upto/client'
Can't resolve '@x402/svm/exact/client'
```

Same defect as #398, reaching ai-chat by a different route — `thirdweb` → `@base-org/account` → `@coinbase/cdp-sdk`, which declares the `@x402` packages as optional peers and then imports them statically. #434 fixed this for `apps/web`; ai-chat has its own `next.config.ts` and never got it.

Applied the same `resolve.alias → false` loop, unchanged in behaviour: only packages that genuinely fail to resolve are stubbed, so installing them turns the stub off by itself. `createRequire(import.meta.url)` since this config is TypeScript rather than the base template's `.js`.

## 3 and 4. Two React 19 typing changes

The template pins `react: ^19.0.0`, and its types moved under it:

- `components/document-preview.tsx` — `useRef<HTMLDivElement>(null)` now yields `RefObject<HTMLDivElement | null>`, while the `hitboxRef` prop was declared `RefObject<HTMLDivElement>`. Widened the prop.
- `components/toolbar.tsx` — `useRef<ReturnType<typeof setTimeout>>()` no longer type-checks without an initial value, so it is `useRef<… | undefined>(undefined)`.

The second file also needed a cast at `useOnClickOutside(toolbarRef, …)`. **That one is not a version problem**: `usehooks-ts@3.1.1` is current and types the parameter `RefObject<T extends HTMLElement>`, with no nullable form, so no bump fixes it. Checked the implementation before casting rather than after — `index.js:535` reads `ref.current && !ref.current.contains(target)`, so it guards the null itself and the cast describes runtime behaviour instead of hiding a hazard. Noted at the call site with a line saying to drop it if upstream widens the signature.

## Verification

Generated a project from the template after the changes, not before:

```
$ celo-composer create fresh -t ai-chat -y --skip-install
$ pnpm install
$ npx tsc --noEmit          # 0 errors
$ npx next build            # exit 0
```

`next build` reaches the route table and exits 0. Note this is `next build` on its own — the package script is `tsx lib/db/migrate && next build`, and the migrate step needs a real `POSTGRES_URL`, which is the separate documented requirement rather than a template defect.

**What is not fixed here:** the deployability items in #387 beyond the build itself. This closes the "does it compile" half; if the issue is tracking more than that, happy to leave it open and retitle.

### DIFF
--- a/templates/ai/chat-template/app/(chat)/layout.tsx
+++ b/templates/ai/chat-template/app/(chat)/layout.tsx
@@ -5,8 +5,6 @@ import { DataStreamProvider } from "@/components/data-stream-provider";
 import { SidebarInset, SidebarProvider } from "@/components/ui/sidebar";
 import Script from "next/script";
 
-export const experimental_ppr = true;
-
 export default async function Layout({
   children,
 }: {
--- a/templates/ai/chat-template/components/document-preview.tsx
+++ b/templates/ai/chat-template/components/document-preview.tsx
@@ -146,7 +146,7 @@ const PureHitboxLayer = ({
   result,
   setArtifact,
 }: {
-  hitboxRef: React.RefObject<HTMLDivElement>;
+  hitboxRef: React.RefObject<HTMLDivElement | null>;
   result: any;
   setArtifact: (
     updaterFn: UIArtifact | ((currentArtifact: UIArtifact) => UIArtifact),
--- a/templates/ai/chat-template/components/toolbar.tsx
+++ b/templates/ai/chat-template/components/toolbar.tsx
@@ -10,6 +10,7 @@ import {
   type Dispatch,
   memo,
   type ReactNode,
+  type RefObject,
   type SetStateAction,
   useEffect,
   useRef,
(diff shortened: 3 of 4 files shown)


GigaHierz (approved):
Verified: ppr removed in both places (config + route-level export, the only two references), the optional-peer stub loop matches #434's base-template pattern, and both React 19 typing fixes check out against the pinned versions. Merging with the body changed from 'Closes #387' to 'Refs #387': the migrate step ('build': 'tsx lib/db/migrate && next build') still hard-fails on a fresh scaffold without POSTGRES_URL, so #387 stays open for that last blocker — a graceful skip-with-warning when the var is unset (keeping migration when set) would let it close for real.

PR #457, opened 2026-08-13
fix(templates): keep the thirdweb scaffold building without a client id

### LINKED ISSUE #452: thirdweb template: fresh scaffold fails `pnpm build` — client.ts throws at module scope when NEXT_PUBLIC_THIRDWEB_CLIENT_ID is unset
## What happens

A fresh thirdweb scaffold fails `pnpm build` before the user has configured anything:

```bash
npx @celo/celo-composer create demo -t basic --wallet-provider thirdweb -c none -y
cd demo && pnpm install && pnpm build
```
```
Error: Missing NEXT_PUBLIC_THIRDWEB_CLIENT_ID environment variable
Error occurred prerendering page "/". Read more: https://nextjs.org/docs/messages/prerender-error
```

With the variable set, the same build passes end-to-end (verified against current main, post-#428/#434).

## Root cause

`templates/wallets/thirdweb/lib/client.ts.hbs` throws at module scope when `NEXT_PUBLIC_THIRDWEB_CLIENT_ID` is unset:

```ts
const clientId = process.env.NEXT_PUBLIC_THIRDWEB_CLIENT_ID!;
if (!clientId) {
  throw new Error("Missing NEXT_PUBLIC_THIRDWEB_CLIENT_ID environment variable");
}
```

The throw executes during Next's static prerendering of every page that transitively imports the client, so the out-of-the-box build is red until the env file is filled in.

## Why this matters / suggested fix

The sibling rainbowkit template handles the identical situation with a placeholder fallback (`|| "YOUR_PROJECT_ID"`), so its fresh scaffold builds green and fails meaningfully only at runtime when a real connection is attempted. The thirdweb template should match: fall back to a placeholder (thirdweb's client creation doesn't validate the id at construction time) or defer the check to runtime with a console warning, keeping the fresh-scaffold build green.

Found while running the post-merge verification matrix on main (all other fixed templates now build clean).

ghozzza:
Closes #452.

Reproduced exactly as described: a fresh `-t basic --wallet-provider thirdweb -c none` scaffold compiles, then fails prerendering `/` and `/_not-found` with `Missing NEXT_PUBLIC_THIRDWEB_CLIENT_ID environment variable`.

## The fix, and the claim it rests on

Took the placeholder route you suggested, matching what rainbowkit already does with `|| 'YOUR_PROJECT_ID'`.

That only works if thirdweb really does not validate the id at construction, so I checked rather than assuming — running against the installed `thirdweb` in a generated project:

| `clientId` | result |
|---|---|
| `"YOUR_CLIENT_ID"` | client created |
| `"placeholder"` | client created |
| `"x"` | client created |
| `""` | **throws** `clientId or secretKey must be provided` |

So any non-empty string is accepted and only the empty string is rejected — the placeholder is safe, and `|| "YOUR_CLIENT_ID"` can never produce the empty case.

## Added a warning as well as the fallback

Your issue offered "placeholder **or** runtime check with a console warning". I took both, because the placeholder alone trades one bad failure for another: the build goes green, the developer deploys, and thirdweb then fails on an authenticated call with an error that never mentions the missing variable.

```
[thirdweb] NEXT_PUBLIC_THIRDWEB_CLIENT_ID is not set, using a placeholder.
Wallet connection will not work until you set it in apps/web/.env.
```

It fires during the build too, since prerendering evaluates the module — which is the right moment to see it.

## Verification

| | result |
|---|---|
| `pnpm build`, no env set | **exit 0**, compiles, warning printed |
| `pnpm build`, `NEXT_PUBLIC_THIRDWEB_CLIENT_ID` set | exit 0, **no warning** |
| CLI suite | 15/15 |

The second row is the one that matters for regression: the fallback and the warning both stay out of the way when the variable is present.

One thing left alone, and worth flagging rather than fixing quietly: `.env.template` ships `NEXT_PUBLIC_THIRDWEB_CLIENT_ID=your_client_id_here`. A developer who copies that file without editing it gets a non-empty value, so no warning fires and the placeholder path is never taken — they reach the same authenticated-call failure with no hint. Rainbowkit's `your_project_id_here` has the identical shape. That is a wider question about whether the env templates should ship empty values rather than filled placeholders, so it did not belong in this fix.

### DIFF
--- a/templates/wallets/thirdweb/lib/client.ts.hbs
+++ b/templates/wallets/thirdweb/lib/client.ts.hbs
@@ -1,12 +1,23 @@
 import { createThirdwebClient } from "thirdweb";
 
-// Replace this with your client ID get one from https://thirdweb.com/dashboard/
-const clientId = process.env.NEXT_PUBLIC_THIRDWEB_CLIENT_ID!;
+// Get a client ID from https://thirdweb.com/dashboard and set it in
+// apps/web/.env as NEXT_PUBLIC_THIRDWEB_CLIENT_ID.
+//
+// This falls back to a placeholder rather than throwing. The page tree imports
+// this module, so a throw here runs during the prerender step of `next build`
+// and makes a fresh scaffold fail before anyone has had a chance to configure
+// anything. `createThirdwebClient` accepts any non-empty string without
+// validating it, so the build stays green and the failure lands where it means
+// something — the first call that actually needs the ID.
+const clientId = process.env.NEXT_PUBLIC_THIRDWEB_CLIENT_ID;
 
 if (!clientId) {
-  throw new Error("Missing NEXT_PUBLIC_THIRDWEB_CLIENT_ID environment variable");
+  console.warn(
+    "[thirdweb] NEXT_PUBLIC_THIRDWEB_CLIENT_ID is not set, using a placeholder. " +
+      "Wallet connection will not work until you set it in apps/web/.env."
+  );
 }
 
 export const client = createThirdwebClient({
-  clientId,
+  clientId: clientId || "YOUR_CLIENT_ID",
 });


GigaHierz (approved):
Verified: matches the rainbowkit placeholder precedent exactly, no module-scope throw so fresh-scaffold prerender succeeds, warning fires at build and runtime, and the empty-string edge is unreachable through ||. Fixes #452.

PR #462, opened 2026-08-13
chore: remove tracked .DS_Store

oksanaphmn:


GigaHierz (approved):
Verified: the root .DS_Store is the only one tracked, and .gitignore already covers .DS_Store since #448, so this can't recur. Thanks for the cleanup!

Issue #470, opened 2026-09-14
Template dependencies in .hbs manifests are invisible to Dependabot and Renovate — a vulnerable pin ships to every scaffold unreported


GigaHierz:
## What happens

`templates/base/apps/web/package.json.hbs` is a dependency manifest that no scanner reads.
It is not named `package.json`, so neither Dependabot nor Renovate has a manager that
matches it, and the dependencies inside it are invisible to both.

The gap is directly observable. Dependabot has alerts open against the one Next.js manifest
in this repo that is a real `package.json`, and none against the Handlebars one — while both
declare a vulnerable `next`:

```
$ gh api "/repos/celo-org/celo-composer/dependabot/alerts?state=open&per_page=100" \
    --jq '[.[]|select(.dependency.package.name=="next")] | .[] | "\(.security_advisory.ghsa_id) \(.security_advisory.severity) \(.dependency.manifest_path)"'
GHSA-2xp9-vwfh-vxw4 critical templates/ai/chat-template/package.json
GHSA-p293-qw3h-jr36 critical templates/ai/chat-template/package.json
GHSA-p9j2-gv94-2wf4 high     templates/ai/chat-template/package.json
...20 more, all templates/ai/chat-template/package.json
```

Zero rows for `templates/base/apps/web/package.json.hbs`, which pins `next: ^14.0.0` —
inside the vulnerable range of both criticals above (`>= 10.0.0, < 15.5.24`) and with no
14.x patch in existence.

Renovate is equally blind. Its Dependency Dashboard (#266) lists no dependency from any
`.hbs` manifest.

## Root cause

`templates/base/apps/web/package.json.hbs:12` — `"next": "^14.0.0"`, in a file whose
extension puts it outside every default manager's `fileMatch`.

The same shape applies to every other `.hbs` manifest the templates ship:

```
$ git ls-files 'templates/**/*.hbs' | grep -i 'package.json'
templates/base/apps/web/package.json.hbs
```

One today. The point is that adding a second creates a second blind spot silently.

## Impact

The generated output is the product. Every project scaffolded by
`npx @celo/celo-composer@latest create` gets its web app dependencies from this file, so
every scaffold inherits whatever has rotted in it — and the rot is never reported, to us or
to the developer who ran the command. A Celo developer starting a new project today gets an
app exposed to an unauthenticated RCE, from a starter kit, with no alert anywhere in the
chain.

This is the reason #458 sat for a month at `priority:high` while the underlying advisories
[...]

Issue #458, opened 2026-08-13
Templates pin Next.js 14, which has no patch for three HIGH advisories — the fix line starts at 15.5.21


ghozzza:
Filed per the 2026-08-12 standup.

## The problem

`templates/base/apps/web/package.json.hbs` pins `next: ^14.0.0`, which resolves to **14.2.35**. Three HIGH-severity advisories cover that version and **none of them has a 14.x patch**:

| advisory | severity | vulnerable | first patched |
|---|---|---|---|
| [GHSA-89xv-2m56-2m9x](https://github.com/advisories/GHSA-89xv-2m56-2m9x) — SSRF in Server Actions on custom servers | HIGH | `>= 14.1.1, < 15.5.21` | **15.5.21** |
| [GHSA-p9j2-gv94-2wf4](https://github.com/advisories/GHSA-p9j2-gv94-2wf4) — SSRF in rewrites via attacker-controlled destination hostname | HIGH | `>= 12.0.0, < 15.5.21` | **15.5.21** |
| [GHSA-m99w-x7hq-7vfj](https://github.com/advisories/GHSA-m99w-x7hq-7vfj) — DoS in App Router using Server Actions | HIGH | `>= 13.0.0, < 15.5.21` | **15.5.21** |

Every fix landed on the 15.5.x line, so there is no patch release to move to inside 14. Staying on 14 means staying exposed, which makes this a version bump rather than a dependency update.

`pnpm audit` on a freshly generated `-t farcaster-miniapp` scaffold reports **41 advisories, 21 of them against `next`** — 8 high, 11 moderate, 2 low. Every `patched_versions` range on those 21 begins at 15.x.

## What the upgrade actually involves

Not as much as it sounds. Tested against a `minipay` scaffold with `next` bumped to `^16.3.0`, `eslint-config-next` to `^16.3.0` and `typescript` to `^5.9.3`:

- **React stays on 18.** Next 16 peers `react: ^18.2.0 || ^19.0.0`, so no React major is forced.
- **`pnpm type-check` passes.**
- **The templates use none of the Next 15 breaking surface.** `grep` for `cookies()`, `headers()`, `params`, `searchParams` and `draftMode()` across `templates/base`, `minipay`, `farcaster-miniapp` and `wallets/` returns nothing — the async request APIs are the main 14→15 break and no template touches them.

[...]

PR #389, opened 2026-06-17
docs: fix typos, broken links, and outdated references

0xward:
Hi team,

I've been going through the documentation and noticed a few small areas that could use some cleanup. I’ve put together this PR to address the following:

- **Fixed typos and formatting:** Corrected duplicate words and restructured the Changelog intro.
- **Foundry fix:** Fixed a broken deployment command that was missing necessary flags.
- **Updated API references:** Updated Etherscan references to Celoscan where appropriate.
- **Content updates:** Updated the Changelog to reflect the transition to Celo Sepolia testnet.

Everything should be straightforward, but please let me know if you'd like any adjustments. Happy to tweak anything!

PR #471, opened 2026-09-14
fix(templates): raise the Next.js floor to 15.5.25 in both template manifests

GigaHierz:
## The hole, and the fix

`templates/base/apps/web/package.json.hbs` pinned `next: ^14.0.0`.
[GHSA-2xp9-vwfh-vxw4](https://github.com/advisories/GHSA-2xp9-vwfh-vxw4) — unauthenticated RCE
in the Image Optimization API when AVIF files are used, **critical** — is vulnerable
`>= 10.0.0, < 15.5.24`. The lower bound is the point: **no 14.x release fixes it**, so the
caret range that looks like it can receive patches cannot receive this one. A range is not
a safety property; only a floor is.

`templates/ai/chat-template/package.json` pinned `15.3.6`, below the same line.

Both now sit at **15.5.25**, the last published 15.x (npm `backport` dist-tag), along with
their `eslint-config-next` pins.

Two design points, both of which are why this is a four-line diff rather than a migration:

- **Staying inside 15 rather than going to 16.** #458 identified the Turbopack default in
  Next 16 as "the whole migration" — `templates/base/apps/web/next.config.js.hbs` carries a
  `webpack:` block holding the `resolve.alias → false` stub for the unresolved `@x402`
  optional peers (#398 / #434), and Next 16 refuses to build with a `webpack` config and no
  `turbopack` config. 15.5.25 clears both criticals and leaves webpack as the default, so
  that block and the equivalent one in the chat template's `next.config.ts` are untouched.
  The build output below still shows `chunks/webpack-*.js`, which is the evidence that the
  bundler did not move.
- **React does not move.** `next@15.5.25` peers `react: ^18.2.0 || ^19.0.0`, so
  `react: ^18.2.0` in the base manifest stays as it is. No React major is forced.

The new test, `src/generators/dependency-floors.test.ts`, sweeps **`package.json.hbs` as
well as `package.json`**. That asymmetry is the actual lesson of this incident: Dependabot
had over twenty alerts open against the chat template's real manifest and **zero** against
the Handlebars one, while both declared a vulnerable `next`. The difference was the file
extension, not the risk. Manifests are discovered by walking `templates/`, not listed, so a
new template that pins its own `next` is caught without anyone remembering to add it.

## What this does NOT do / residual risk

- **It does not make a generated project advisory-free.** On a `farcaster-miniapp`
  scaffold, `pnpm audit` goes from the **41 advisories (21 against `next`)** recorded in
  #458 to **21 advisories, 0 against `next`**. The remaining 21 are the wallet stack, not
  the framework. Numbers below.
- **It does not make the templates scannable.** The `.hbs` blind spot itself is #470 — a
  Renovate `customManagers` entry. This PR's floor test is a backstop for that, not a
  replacement: the test enforces a floor someone has to remember to raise, while the
  Renovate manager would raise it unprompted. Both are wanted.
- **It does not help until Renovate is actually running here.** The last Renovate PR on this
  repo is #352, dated 2025-04-04, despite the Dependency Dashboard (#266) existing. Tracked
  outside this repo.
- **`sharp` is fixed as a side effect, not by intent.** `next@15.5.21` and below declare
  `sharp: ^0.34.3`, which tops out at `0.34.5` and so made the patched `0.35.4`
  (GHSA-rgj7-g3m4-5g8c, libheif AVIF heap overflow) genuinely unreachable. `15.5.25`
  declares `^0.34.3 || ^0.35.4`. Every scaffold below resolves `sharp@0.35.4` with no
  override. Worth knowing because it inverts the obvious order of work — the `sharp`
  advisory could not have been fixed first.
- **The floor test reads exact pins and caret/tilde ranges only.** A manifest using another
  range shape throws with a message telling the reader to teach it that shape rather than
  drop the manifest from the sweep — a silently-skipped manifest is the failure mode this
  whole suite exists to prevent.

## Judgement calls

**Both templates in one PR, rather than two.** #458 scopes both manifests as one fix unit,
the floor test spans both, and splitting would mean a test that is red on `main` between
the two merges. One concern — "raise the Next.js floor above the advisory line" — expressed
in two files.

**15.5.25, not 15.5.24.** The advisory floor is 15.5.24; 15.5.25 is the newest release on
the line and includes it. The test asserts `>= 15.5.24`, the advisory's number, so the
constant that matters is the advisory's rather than today's latest.

**The test lives in `src/generators/`, beside the existing suites,** even though it is not
about generation. It is where the scaffold-driving helpers already are, and the second
describe block does drive a real `create`.

**Not bundled:** no product change, no `next.config` change, no React bump, no dependency
touched other than `next` / `eslint-config-next`.

## Issues

Refs #458 — closes the "Revised scope" bullets for both manifests. Leaves open the ops box
about confirming alert state, which only Dependabot can close, and #458's own discussion of
the 16 upgrade as a separate future decision.
Refs #470 — related; #470's `customManagers` diff is not in here.

## Stacking / conflicts

Branched off `main` @ b1eeba3, independent. No shared files with any other open PR (#468 is
a template-sync PR touching `templates/` metadata, not these manifests; #389 is docs).

## Verification evidence

**Mutation count: reverting the two manifest edits turns 3 tests red.**

```
$ npx jest src/generators/dependency-floors.test.ts     # with the manifests reverted
 ● templates/ai/chat-template/package.json declares next >= 15.5.24
     -   "atLeast": true
     +   "atLeast": false,
         "next": "15.3.6",
 ● templates/base/apps/web/package.json.hbs declares next >= 15.5.24
     +   "atLeast": false,
         "next": "^14.0.0",
 ● a scaffolded project ships a patched next › declares next >= 15.5.24 in the rendered apps/web manifest
     +   "atLeast": false,
         "next": "^14.0.0",

Tests:       3 failed, 3 passed, 6 total
```

**On this head:**

```
$ pnpm test
PASS src/generators/dependency-floors.test.ts
PASS src/generators/template-wiring.test.ts (9.983 s)
PASS src/generators/scaffold-smoke.test.ts (10.809 s)
Tests:       34 passed, 34 total

$ pnpm lint
✖ 1 problem (0 errors, 1 warning)      # the warning is pre-existing, in scaffold-smoke.test.ts:144

$ pnpm build
> tsc                                   # exit 0
```

**The real artifact — four scaffolds generated by the CLI, installed and built.** Per the
repo's own gotcha that the CLI building is not enough:

```
==================== basic-rk (-t basic --wallet-provider rainbowkit) ====================
-- declared next: ^15.5.25
-- resolved next: next@15.5.25
-- resolved sharp: sharp@0.35.4
basic-rk BUILD OK
==================== basic-tw (-t basic --wallet-provider thirdweb) ====================
-- declared next: ^15.5.25   -- resolved next: next@15.5.25   -- resolved sharp: sharp@0.35.4
basic-tw BUILD OK
==================== mp (-t minipay) ====================
-- declared next: ^15.5.25   -- resolved next: next@15.5.25   -- resolved sharp: sharp@0.35.4
mp BUILD OK
==================== fc (-t farcaster-miniapp) ====================
-- declared next: ^15.5.25   -- resolved next: next@15.5.25   -- resolved sharp: sharp@0.35.4
fc BUILD OK
```

`declared` is read out of the rendered `apps/web/package.json`, so it also confirms the
`.hbs` edit renders valid JSON around the conditional wallet blocks — the failure mode where
an edited Handlebars manifest reads fine and parses as nothing.

**Advisory count on a generated project, like-for-like with the measurement in #458:**

```
$ cd fc && pnpm audit --json | ...
total: 21 | next: 0 | sharp: 0
by severity: {'moderate': 16, 'low': 1, 'high': 4}
```

#458 recorded `41 advisories, 21 of them against next — 8 high, 11 moderate, 2 low` on the
same `-t farcaster-miniapp` scaffold. 41 → 21, and all 21 `next` advisories are gone.

**The AI chat template builds:**

```
$ ./node_modules/.bin/next build
Route (app)                                 Size  First Load JS
┌ ƒ /                                      186 B        1.17 MB
...
+ First Load JS shared by all             117 kB
  ├ chunks/webpack-8c2772c2380f8065.js   16.5 kB      ← still webpack, not Turbopack
ƒ Middleware                             82.8 kB
```

`pnpm build` in that template runs `tsx lib/db/migrate` first and needs `POSTGRES_URL`, so
`next build` was invoked directly. The framework bump is what is under test here; the
migration step is unrelated to it.

Resolution in that template: `next@15.5.25`, `sharp@0.35.4`.

**Security note worth the maintainer's attention.** Of everything in this repo, the AI chat
template is the closest to actually exploitable, and it is worth saying why rather than
leaving it at "critical severity". It ships `app/(chat)/api/files/upload/route.ts`, which
accepts `image/jpeg` and `image/png` — validated from the client-declared `file.type`, which
is spoofable — so an AVIF can be uploaded under a JPEG content type. Two things stop it
today: the upload lands on Vercel Blob, a different origin, and `next.config.ts` allowlists
only `avatar.vercel.sh` in `images.remotePatterns`, so the optimizer will not fetch it.

That is a starter kit. A developer who adds their own blob host to `remotePatterns` — an
entirely ordinary thing to do — removes the only thing standing between the upload path and
the optimizer, on a version where that is an unauthenticated RCE. Patching the template is
the fix; tightening the MIME check to sniff bytes rather than trust `file.type` is a
separate, smaller question I have deliberately not bundled here. Happy to file it.

## Remaining ops steps

- [ ] Confirm the Dependabot alerts on `templates/ai/chat-template/package.json` close after
      merge: `gh api "/repos/celo-org/celo-composer/dependabot/alerts?state=open" --jq '[.[]|select(.dependency.package.name=="next")]|length'` → `0`

## Checklist

- [x] Title is the commit message I want on `main`
- [x] Ships the test that fails on pre-fix code (3 red); fixtures are real generated projects
- [x] Covers the seam — the second describe block drives the real `create` as a subprocess
      and asserts on the rendered manifest, not on the template source
- [ ] Error responses audited — n/a, no request path changed
- [x] Judgement calls flagged above
- [x] `lint` / `build` / `test` pass locally on this head
- [x] Re-read #458's revised scope — both manifest bullets met; ops box left open above
- [x] Development sidebar links match Refs above
- [ ] README / docs — not updated; no user-facing instruction changes. Flagging in case the
      template matrix in `docs/` states a Next version I have not found.
- [x] Lockfile — none touched. The templates ship no lockfile; `pnpm-lock.yaml` at the repo
      root is the CLI's own and is unchanged.
- [x] Wallet/provider tree touched → both `rainbowkit` and `thirdweb` scaffolds built above.
      Not loaded in a browser: this changes a version pin, not wallet code, and all four
      scaffolds compile. Say the word if you want a click-through.
- [x] Security path → patching a critical advisory; exposure analysis above
- [x] No secrets in the diff

## Questions for the maintainer

1. **15 now, or 16 as part of this?** I took 15.5.25 because it clears both criticals
   without the Turbopack decision #458 identifies. That leaves the templates a major behind,
   which #458 also flags. My read is that a critical RCE should not wait on a bundler
   migration — but the 16 move now has no security urgency behind it, which may make it
   harder to schedule, not easier. Worth deciding deliberately rather than by default.
2. Should I file the byte-sniffing MIME check on the chat template's upload route?


ghozzza (commented):
Comment only — I have read access here, so this can't be an approval — but I would approve it. The floor is the right fix and it is urgent; nothing below should hold the merge. Two corrections to the body, one follow-up that decides whether the fix reaches anyone, and two gaps in the new test.

Verified: both advisories are critical, published 2026-09-08, fixed in 15.5.24; 15.5.25 is the last 15.x (`backport` tag); `next@15.5.25` peers `react ^18.2.0 || ^19.0.0` and allows `sharp ^0.35.4`; the repo's only tracked lockfile is the root `pnpm-lock.yaml`, which has no `next`, and the chat template is copied raw, so there is no lockfile holding the old version; no other `next` pin anywhere in the repo. Suite 28 → 34 (+6, all in the new file). Mutations: both manifests reverted → 3 red, the `.hbs` alone → 2, a new template pinning 15.3.6 → 1, `"latest"` → 2 with the intended error, `">=14.0.0 <16"` → red.

**1. Merging does not protect a single user until a release ships (Medium).** `npx @celo/celo-composer@latest create` installs from npm, and npm `latest` is still **2.4.13** (published 2026-03-09), whose tarball carries `"next": "^14.0.0"` in `templates/base/apps/web/package.json.hbs` and `"next": "15.3.0-canary.31"` in the chat template. The repo's `package.json` is also 2.4.13. So every scaffold stays vulnerable after this merges, until a patch release is published — worth one line under "What this does NOT do" and a named owner for the publish.

**2. "React does not move" is true of the manifest, not of what runs (Medium).** The base template is App Router (`templates/base/apps/web/src/app/layout.tsx.hbs`), and Next aliases `react$`/`react-dom$` to its own vendored copy for the app directory (`next@15.5.25`, `dist/build/create-compiler-aliases.js:251-258`): 14.2.35 vendors a React 18.3 canary, 15.5.25 vendors a **19.2 canary**. So RainbowKit and thirdweb render on React 19 at runtime even though the manifest keeps `react ^18.2.0`. I read that from Next's source and did not observe it in a browser, but it takes away the body's reason for skipping a click-through ("changes a version pin, not wallet code"). One connect-wallet per provider on a fresh scaffold would settle it.

**3. The floor test reads only the first `"next"` in a manifest (Low).** `dependency-floors.test.ts:97` uses `source.match`, so a second declaration is invisible: adding `pnpm.overrides.next = "14.2.35"` to the chat template, or a second `{{#if …}}"next": "^14.0.0"{{/if}}` in the `.hbs`, both leave it 6/6 green. `matchAll`, asserting every match.

**4. The parser accepts a prerelease tail (Low).** `:43-44` strips the prefix and `parseInt`s each part, so `15.5.24-canary` reads as 15.5.24 and passes the floor although it sorts below it; `>=A <B` is also accepted rather than thrown (harmless, since the lower bound is read correctly), so "another range shape throws" is only partly true. Refusing anything that does not match `^[\^~]?\d+\.\d+\.\d+$` closes both.

On the audit numbers: a `farcaster-miniapp` scaffold today goes from 35 advisories (23 against `next`, 2 critical) at the merge-base to 11 (0 against `next`) at the head. Same direction as the body's 41/21 → 21/0; the absolute counts move with the advisory database, so the "0 against next" half is the one that matters.

Not this PR, noticed on the way: `"lint": "next lint"` now prints a deprecation for Next 16 on every scaffold, and the base `engines` (`>=18.0.0`, `templates/base/package.json.hbs:38`) is below what Next requires (`^18.18.0`) — already true on 14.2.35.

ghozzza:
Two corrections to my review above. **2.4.13 was published on 2025-12-18**, not 2026-03-09 — that date is npm's `time.modified`, not the release. And Next's full engines range is `^18.18.0 || ^19.8.0 || >=20.0.0`. Neither changes the point: `latest` is still 2.4.13, and it still ships `"next": "^14.0.0"` in the base template.

